PR #198, opened 2018-08-28
Clean up

markt-asf:
Clean-up of warnings after opening code in Eclipse.
No functional changes (so should be safe to include in next release)
I find having a clean (i.e. no warnings) code base makes it easier to spot errors such as JavaDoc errors in future development work.
The copyright header has been updated as per Eclipse guidance for all modified files.
There is scope for further clean-up - such as consistent formatting - but that needs a community discussion to agree what we want that format to be first.

### DIFF
--- a/src/main/java/javax/servlet/GenericFilter.java
+++ b/src/main/java/javax/servlet/GenericFilter.java
@@ -16,7 +16,6 @@
 
 package javax.servlet;
 
-import java.io.IOException;
 import java.util.Enumeration;
 import java.util.ResourceBundle;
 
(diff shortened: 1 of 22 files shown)


arjantijms:
Big +1 for all these changes.

As for the code formatting, there's an EE4J guideline: https://github.com/eclipse-ee4j/ee4j/wiki/Code-Conventions

It's essentially the well known Sun conventions, with a few small exceptions. This is relatively easy to create in Eclipse. 

Starting with the "Eclipse [build-in]" formatter, basically change tab policy to "spaces only" and indentation and tab size to 4. Under line wrapping change maximum line width to 160. And under comments maximum line width to 120. Under Javadocs, remove "indent wrapped @param" and "new line before @param"

arjantijms:
ps max line at 160 means that the max is 160 for cases that warrant it. In most cases I think lines should not be much wider than ~120.

markt-asf:
Tx. I'll start a thread on the servlet-dev list about applying that standard formatting.

PR #202, opened 2018-09-04
Format part of fix for #126 that wasn't reformatted on merge

markt-asf:
There was a conflict between the large reformat fix and the Javadoc updates. This text wasn't in conflict so the format wasn't updated.


m0mus:
You will possibly need to backport it into EE4J_8 branch.

markt-asf:
That was going to be my next PR ;)

PR #203, opened 2018-09-04
Merge clean-up, formatting and doc fixes to EE4J_8 release branch

markt-asf:
All of these changes should be safe for the EE4J 8 release. The public API remains unchanged both in terms of method signatures and expected behaviour as described in the Javadoc.


pom.xml
```diff
@@ -152,6 +152,9 @@
                 <configuration>
                     <archive>
                         <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                        <manifestEntries>
+                            <Automatic-Module-Name>java.servlet</Automatic-Module-Name>
```
gregw: This change should not be included with doc and formatting changes

gregw: plus it is subject to another PR

src/main/java/javax/servlet/AsyncContext.java
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2018 Oracle and/or its affiliates and others.
+ * All rights reserved.
```
gregw: Typically I'm cautious about reformatting such copyright lines, as they often are targets of tools to update etc.
On what basis have these lines been split? There are longer lines elsewhere in comments?

markt-asf:
Note that this PR is not for new changes. It is simply to merge all of the changes that have already been applied to master to the EE4J_8 branch.

Addressing the individual comments
- The JPMS name change is in a separate commit. It just happens to be in amongst other reformatting changes.
- It is too late to add the formatting details to the commit but we can add them to the CONTRIBUTING.md file we are going to have to add before a release.
- The license header was part personal taste, part experience in WebSocket that has a plug-in checking the headers. The license header is wrapped at 80 characters so `All rights reserved.` was going to be split across a line. The exact location of the split is going to depend on how the year information is added. If we start down the 2015, 2016-2018 route (I don't think we should but that is a separate discussion) the split will vary between files breaking the header check. Moving `All rights reserved.` to a new line was a simple way to avoid any (potential) issues.

stuartwdouglas:
I don't think we should be merging this. According to 'PMC statement about initial EE4J project contributions' at https://www.eclipse.org/ee4j/news/?date=2018-02-16

There are two bullet points that are problematic:

> In order to merge a PR into the EE4J_8 branch project lead or PMC approval is required and the change must be tested against the TCK which is currently not available.

and

> The PMC strongly recommends accepting only critical bug fixes to the EE4J_8 branch.

I don't think that we can really count the reformatting as a 'critical bug fix' so it should probably wait till the next platform release.

The pom.xml changes on the other hand we will obviously need, but in theory we should run the TCK against them.

markt-asf:
Things have moved on a bit since this PR was submitted. I'm going to close this one. I expect I'll submit some smaller ones after some discussion on the dev@ list

PR #204, opened 2018-09-06
Fixes #201 - Java JPMS modules.

### LINKED ISSUE #201: Java JPMS modules
The servlet API artefact should be consumable as a JPMS module for java 11

sbordet:
Added Automatic-Module-Name: javax.servlet.api to the MANIFEST.

Signed-off-by: Simone Bordet <simone.bordet@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,9 @@
                 <configuration>
                     <archive>
                         <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                        <manifestEntries>
+                            <Automatic-Module-Name>javax.servlet.api</Automatic-Module-Name>
+                        </manifestEntries>
                     </archive>
                     <excludes>
                         <exclude>**/*.java</exclude>


ggam:
My understanding is that the module name should be `java.servlet`. See JSON-P for example: https://github.com/eclipse-ee4j/jsonp/blob/master/api/src/main/jdk9/module-info.java

markt-asf:
Should be on what basis? This name was chosen to align (as closely as possible) to the OSGI bundle name of "javax.servlet-api". (See #201)

sbordet:
I don't have a personal preference, for me both are valid. About the link with OSGi, it's true that the symbolic name resemble the module name, but that's about it (and they are different anyway).

I'd be happy to update the PR if we reach a consensus on a different module name.

It will be important that they are consistent though.

We don't want the Servlet module to be the only one that has `.api` if the others don't.

E.g. what is the proposed module name for CDI, JSP, JPA, etc.?

ggam:
I can't find an official document, but I recall there was some discussion about this during Java EE 8 development, and the Oracle recommendation was to use "java" instead of "javax" for module names. The reason is that there's no need to distinguish "core" vs "extension" modules anymore, which was the reason to add the "x" suffix. All Java SE modules, even the ones with "javax" packages, are named "java.*".

In addition to this, and given that package names must be unique between modules, the general recommendation is to use the top level package as the module name.

The last published versions of JSON-P, JSON-B, JAX-RS, JPA and I think JTA and JAXB already adopted this convention.

jodastephen:
I believe I've seen something official that says JCP spec modules should use `java.xxx` module naming (as they are special, and distinct from my more general naming recommendations). I can't find the link though...

bshannon:
It should be java.servlet.

(Really, it should be decided by the Jakarta EE Specification Committee
as a part of Jakarta EE 9, but hopefully they would agree that it should
be java.servlet.  Definitely not javax.servlet.)

sbordet:
I have updated the PR to `java.servlet`.

PR #210, opened 2018-10-06
Remove maven-remote-resources-plugin

stuartwdouglas:
This plugin was adding the CCDL license to the final jar


arjantijms:
Good catch!

I'm okay with this change, but just wondering if there aren't any Eclipse license coordinates to use instead.

arjantijms:
The plugin loads a single file from maven central (the license file) and adds that to the final jar.

PR #218, opened 2018-10-24
All of these files have been modified since the initial contribution.

markt-asf:
Therefore the license header needs to be updated. All of them became
longer than 80 chars. I opted to break the line in the same place as it
will make any future automated license header checking simpler to
implement (based on my experience of this in the WebSocket project).

Signed-off-by: Mark Thomas <markt@apache.org>


arjantijms:
Thanks @gregw :)

PR #234, opened 2019-01-09
Fix #233 Update getMapping javadoc and associated clean-up

### LINKED ISSUE #233: Fix javax.servlet.http.HttpServletMapping.getMapping javadoc about return value when using DEFAULT servlet
[HttpServletMapping](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletMapping.html) getPattern method states the following about DEFAULT:

> If the getMappingMatch value is CONTEXT_ROOT or DEFAULT, this method must return the empty string

When reading the same javadoc few lines before, there is a table where slash pattern is written:

> URI Path (in quotes) | matchValue | pattern | mappingMatch
> -- | -- | -- | --
> "" | "" | "" | CONTEXT_ROOT
> "/index.html" | "" | / | DEFAULT
> "/MyServlet" | MyServlet | /MyServlet | EXACT
> "/foo.extension" | foo | *.extension | EXTENSION
> "/path/foo" | foo | /path/* | PATH
> 

Additionally, In servlet 4 spec., section 12.2 says:

> A string containing only the ’ / ’ character indicates the "default" servlet of the
application

So, this collides with HttpServletMapping.getPattern statement.

If both values (empty and "/") are valid for DEFAULT servlet, then this should be written in method's javadoc.

markt-asf:


src/main/java/javax/servlet/http/HttpServletMapping.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates and others.
```
gregw: Should we just do a "Happy New Year" commit and update all the headers?

bshannon: Please don't...

src/main/java/javax/servlet/http/HttpServletMapping.java
```diff
@@ -97,13 +97,13 @@
...
      * <p>
-     * Return the portion of the URI path that caused this request to be matched. If the {@link #getMappingMatch} value
-     * is {@code
-     * CONTEXT_ROOT} or {@code DEFAULT}, this method must return the empty string. If the {@link #getMappingMatch} value
-     * is {@code
-     * EXACT}, this method must return the portion of the path that matched the servlet, omitting any leading slash. If
-     * the {@link #getMappingMatch} value is {@code EXTENSION} or {@code PATH}, this method must return the value that
-     * matched the '*'. See the class javadoc for examples.
+     * Return the portion of the URI path that caused this request to be matched. If the {@link #getMappingMatch} value is
+     * {@code
```
stuartwdouglas: This formatting still looks a bit weird, should there actually be a line break here?

src/main/java/javax/servlet/http/HttpServletMapping.java
gregw: @bshannon is there a specific jakarta policy that copyright year will be updated at a file level granularity?   I would have thought that if we release a new version of the API, then the copyright year will be updated for the whole API and not just the files that have been changed?

src/main/java/javax/servlet/http/HttpServletMapping.java
markt-asf: Ideally, there should not. That reformatting was the result of applying the standard EE4J formatting ruleset for the Eclipse IDE to the file. I noticed the extra line breaks but opted not to fix them. On one hand, it doesn't look quite right in the source code. On the other it won't affect the look of the Javadoc the user sees and manually 'fixing' the formatting after applying an automated tool (and having to fix them after every future reformatting) seemed a little pointless.

Issue #243, opened 2019-07-11
Prepare Jakarta Servlet for Jakarta EE 8 Release


m0mus:
### Tasks

- [x] Update text files
- [x] Generate "boilerplate" specification project
- [x] Make a staging release
- [x] Generate Stand-alone TCK test results
- [x] Submit a PR to jakartaee/specifications repository
- [x] Update the Jakarta EE API jar by submitting a PR to the [jakartaee-api](https://github.com/eclipse-ee4j/jakartaee-api) project

See [here](https://wiki.eclipse.org/How_to_Prepare_API_Projects_to_Jakarta_EE_8_Release) for detailed instructions.

arjantijms:
[API PR](https://github.com/eclipse-ee4j/jakartaee-api/pull/19)

arjantijms:
[Github pages requested](https://bugs.eclipse.org/bugs/show_bug.cgi?id=549725)

Issue #239, opened 2019-04-04
Update Scope Statement for Jakarta Servlet Project


ivargrimstad:
Update the scope statement for the project: https://projects.eclipse.org/projects/ee4j.jakartaee-platform/governance

See https://docs.google.com/spreadsheets/d/1FPLjo_DJDOQngVtdrYgaDk98d9NBG-0ODgdmOG3LOsI/edit?usp=sharing for examples.

chkal:
Draft:

> Jakarta Servlet provides a specification document, API and TCK for interacting with the web on the Jakarta EE platform.

Based on the wording in the original [project proposal](https://projects.eclipse.org/projects/ee4j.servlet/reviews/creation-review).

chkal:
Any feedback on this proposal?

arjantijms:
Perhaps this one needs a little tuning, although it's a good start.

Interacting with the web, in essence, is what JAX-RS, JAX-WS, WebSocket, JSF and MVC are also doing. 

We probably don't want the Servlet scope to be so broadly stated that it could effectively become JSF (this is of course very far fetched, but just for an extreme example).

Ideally the Servlet scope would explain in a concise wording that's it's a "thin" usability layer on top of HTTP. So its API is higher level than Grizzly or Netty, which are more bare HTTP and other protocol engines, but lower level than JAX-RS.

For better or worse, the Servlet spec also implies implementations having this HTTP engine on top of which its a usability layer, and it contains the core concepts of deploying war achives, and also has some of the security fundamentals. The latter have been formalised in among others the JACC spec, but have not been moved to there, so have always remained in Servlet as well.

So given all of the above, "interacting with the web" might not fully explain what Servlet does.

stuartwdouglas:
I prefer the one in the example spreadsheet:

>Jakarta Servlet provides a specification document, API, and TCK that defines a Java-based means of implementing the HTTP request/response paradigm over network services.

With some minor tweaks:

>Jakarta Servlet provides a specification document, API, and TCK that defines a means of implementing the HTTP request/response paradigm

I don't think we need to include everything that is in Servlet as part of the statement, but I also don't really like 'interacting with the web' as this could also be taken to include websockets, SSE etc.

waynebeaton:
If we boil this down to just what the specification itself defines, we're left with this:

> Jakarta Servlet defines a means of implementing the HTTP request/response paradigm.

That actually feels like enough to me.

[...]

PR #247, opened 2019-07-13
#243 Prepare project for Jakarta EE 8 Release

arjantijms:


api/pom.xml
```diff
@@ -23,22 +23,27 @@
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
         <version>1.0.5</version>
+        <relativePath/>
     </parent>
 
     <groupId>jakarta.servlet</groupId>
     <artifactId>jakarta.servlet-api</artifactId>
-    <version>4.0.2-SNAPSHOT</version>
+    <version>4.0.3-SNAPSHOT</version>
```
gregw: Version number increase should have been done as part of the release process. Bad form to mix it in with substantive changes

README.md
```diff
@@ -13,10 +16,18 @@ Run the build:
...
 Checking findbugs
 -----------------
 
 `mvn -DskipTests -Dfindbugs.threshold=Low findbugs:findbugs`
 
+About Jakarta Servlet
+-------------
+
+Jakarta Servlet defines a server-side API for handling HTTP requests and responses,
+and a packaging model for such server-side applications.
```
gregw: The decision from #239  was that this should just be:

> Jakarta Servlet defines server-side handling for HTTP requests and responses.

api/src/main/java/javax/servlet/AsyncContext.java
```diff
@@ -36,84 +36,82 @@
...
+ * @since 3.0
  */
 public interface AsyncContext {
 
     /**
      * The name of the request attribute under which the original request URI is made available to the target of a
      * {@link #dispatch(String)} or {@link #dispatch(ServletContext,String)}
      */
-    static final String ASYNC_REQUEST_URI = "javax.servlet.async.request_uri";
+    String ASYNC_REQUEST_URI = "javax.servlet.async.request_uri";
```
gregw: These are good changes, but I'm not so sure they should be included in a PR that is substantially about establishing the new project name.

api/pom.xml
arjantijms: I was doubting whether this should be done here, I'll revert it then.

README.md
arjantijms: I think this is just the description, not the official scope. It can go beyond what the scope says, as it's just informative to the one reading what the project is about.

api/pom.xml
arjantijms: Okay, reverted.

arjantijms:
> A general reformatting and cleanup of the code.

Note that reformatting is mostly done to reformat the Javadoc after the abbreviations have been replaced by the full Jakarta names, and the lines on which the replacement was done would stick out. 

Also the reformatting is to align the Javadoc with the Jakarta formatting rules, so I feel that was in scope for a PR about "Prepare project for Jakarta EE 8 Release". If if was just for the name change this might be out of scope indeed, but #243 seemed sufficiently broad.

gregw:
@arjantijms I don't mind the reformatting and the general cleanup of unneeded public, updated since tags etc.      But the resulting changes to 72 files hides the substantive change of the project renaming.    Sure scope for #243 is broad, but that doesn't mean that all the changes have to be in a single PR... or indeed a single commit.       I know that in the early days of this project a lot of short cuts were taken just to get it established, but I think we should now start putting the brakes on bad practises.   Mixing substantive changes (such as renaming the project) with a wider reformatting/cleanup of the code is just bad practise.

README.md
gregw: I think we should just stick with the official scope, at least until we've had a discussion about what we want to say here.

arjantijms:
@gregw I hear you, but the formatting and update of since tags were all part of the rename. Since the project is not called “Servlet” the since tag would have either needed to say “-@since Jakarta Servlet 3.0” or “@since 3.0”. I opted for the latter.

The reformatting or the javadoc is after the expansion of the acronyms as per the referenced issue.

The only “extra” bits were the redundant “public static final” and “public” modifiers. I seem to recall we said to rely on defaults and omit redundant syntax, but tbh that was from memory only.

gregw:
@arjantijms I still think we can do the renaming in a single simple commit with the following sequence:
1. commit to remove the "Servlet" from all the since tags - that is simple and straight forward to approve because the project name should not be there.
2. commit to the project renaming so we can see exactly what has been renamed
3. commit to reformat, cleanup etc.

I think 1 & 2 can be in the same PR, so long as we don't squash.    If 3. is minimal, perhaps it's not even needed at this stage?

arjantijms:
@gregw sure, I’m not against it. If you want, go ahead and close this PR, maybe using it for inspiration or double check for the next PRs/commits you proposed.

PR #248, opened 2019-07-18
Updated format of since tags

gregw:
The since tag format should only contain a version number.


pzygielo:
I'm confused: this vs. https://github.com/eclipse-ee4j/jta-api/pull/46

stuartwdouglas:
This goes against https://wiki.eclipse.org/How_to_Prepare_API_Projects_to_Jakarta_EE_8_Release

"Leave existing "@since XYZ 1.x" uses alone. These refer to the old JCP version. Future additions should use the Jakarta project name."

gregw:
As the XYZ was not strictly needed in these tags, this was an attempt to avoid the issue.  But I'm happy to leave them as is.

PR #249, opened 2019-07-18
Update to version 4.0.3-SNAPSHOT

gregw:
Wasn't done as part of last release.


gregw:
Ooops brain dead.  I don't like forced pushes so will close and reopen another PR

PR #260, opened 2019-08-12
Add doc-files/speclicense.html

stuartwdouglas:


stuartwdouglas:
Actually it probably does. According to the checklist there should be a link to this in the footer of the javadoc.

bshannon:
> I think we should also fix the vendor name in the poms - edit: this will be in a different PR

Note that the vendor for the source code is different than the vendor for the specification.

stuartwdouglas:
I opened https://github.com/eclipse-ee4j/servlet-api/pull/261 with this change but I am not 100% sure what the correct values should be.

Issue #231, opened 2018-12-19
HttpServlet doGet()、doPost()、doPut()、doDelete() etc. methods can refactor


ykgarfield:
Now, HttpServlet doGet()、doPost()、doPut()、doDelete() etc. methods  looks like basically the same:

```
String protocol = req.getProtocol();
String msg = lStrings.getString("http.method_delete_not_supported");
if (protocol.endsWith("1.1")) {
	resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
} else {
	resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
}
```

Except the `lStrings.getString(...)` method param, these methods can extract a method for reuse.

markt-asf:
Fixed in #321

PR #266, opened 2019-09-07
HttpServlet doGet(), doPost(), doPut(), doDelete() etc  methods can refactor #231

ahmadshalabi:
Signed-off-by: ahmadshalabi <eng.ahmadshalabi@gmail.com>

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/LocalStrings_es.properties
+++ b/api/src/main/java/jakarta/servlet/http/LocalStrings_es.properties
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
+# Copyright (c) 1997, 2022 Oracle and/or its affiliates and others.
 # All rights reserved.
 # Copyright 2004 The Apache Software Foundation
 #
@@ -16,8 +16,6 @@
 # limitations under the License.
 #
 
-# $Id: LocalStrings_es.properties,v 1.1.1.1 2003/01/27 16:07:47 ja120174 Exp $
-#
 # Default localized string information
 # Localized para Locale es_ES
 


api/src/main/java/javax/servlet/http/HttpServlet.java
```diff
@@ -137,8 +137,15 @@ public HttpServlet() {
...
+    }
+
+    // resolve method_not_supported message at runtime
+    private void doMethod(HttpServletRequest req, HttpServletResponse resp) throws IOException {
         String protocol = req.getProtocol();
-        String msg = lStrings.getString("http.method_get_not_supported");
+        String errMsg = lStrings.getString("http.method_not_supported");
+        String method = req.getMethod();
+        String msg = MessageFormat.format(errMsg, method);
         if (protocol.endsWith("1.1")) {
```
gregw: Actually this logic is now not so good with HTTP/2.  It is probably only HTTP/1.0 that BAD_REQUEST should be sent and all others should be METHOD_NOT_ALLOWED.... Actaully I could argue that we could just always send METHOD_NOT_ALLOWED as while that was not defined in RFC1945, it was still a meaningful 4xx response code?  @stuartwdouglas your thoughts?

api/src/main/java/javax/servlet/http/HttpServlet.java
```diff
@@ -367,76 +356,44 @@ protected void doDelete(HttpServletRequest req, HttpServletResponse resp) throws
...
-            if (allow.length() > 0) {
-                allow.append(", ");
-            }
-            allow.append(METHOD_TRACE);
-        }
-        if (ALLOW_OPTIONS) {
-            if (allow.length() > 0) {
-                allow.append(", ");
+
+        for (Method value : methods) {
```
gregw: We should calculate this automatic options response either lazily or when the servlet is constructed.  Calculating it on every option request is inefficient.

api/src/main/java/javax/servlet/http/HttpServlet.java
```diff
@@ -367,76 +356,44 @@ protected void doDelete(HttpServletRequest req, HttpServletResponse resp) throws
...
         }
 
+        // TRACE method is allowed since it used in debugging
+        appendAllowedMethod(allow, METHOD_TRACE);
+        appendAllowedMethod(allow, METHOD_OPTIONS);
+
         resp.setHeader("Allow", allow.toString());
     }
 
+    private void appendAllowedMethod(StringBuilder allow, String method) {
```
gregw: this should be static

api/src/main/java/javax/servlet/http/HttpServlet.java
stuartwdouglas: I would just change it so that only HTTP/1.0 sends BAD_REQUEST. I don't think we should be modifying the existing 1.0 behaviour, its probably harmless but it's possible it could cause problems for someone.

gregw: Note that there are probably some HTTP/0.9 interactions still happening... so we should also keep BAD_REQUEST for them too?

stuartwdouglas: I guess so. I don't feel strongly about this either way, but my thinking is keeping the existing behaviour is definitely fine, while changing it could cause problems (even though it is unlikely).

api/src/main/java/javax/servlet/http/HttpServlet.java
```diff
@@ -137,8 +137,15 @@ public HttpServlet() {
      * @see javax.servlet.ServletResponse#setContentType
      */
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        doMethod(req, resp);
+    }
+
+    // resolve method_not_supported message at runtime
+    private void doMethod(HttpServletRequest req, HttpServletResponse resp) throws IOException {
         String protocol = req.getProtocol();
```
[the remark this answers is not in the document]
ahmadshalabi: I prefer to go with moving it before it is used since it more readable.
So, what you think?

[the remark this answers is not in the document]
gregw: Sure, although I'm thinking that perhaps a switch statement is really what is needed below.  See comment there.

api/src/main/java/javax/servlet/http/HttpServlet.java
gregw: How about:
```java
  switch(req.getProtocol())
  {
    case "HTTP/0.9":
    case "HTTP/1.1":
      resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
      break;
    default:
      resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
  }
```

gregw:
#231 #267

api/src/main/java/javax/servlet/http/HttpServlet.java
```diff
@@ -367,76 +356,44 @@ protected void doDelete(HttpServletRequest req, HttpServletResponse resp) throws
...
+                    break;
+                case "doDelete":
+                    appendAllowedMethod(allow, METHOD_DELETE);
+                    break;
             }
-            allow.append(METHOD_OPTIONS);
         }
 
+        // TRACE method is allowed since it used in debugging
+        appendAllowedMethod(allow, METHOD_TRACE);
```
joakime, line 384: I wish there was a way to not have `TRACE` show up in the default `OPTIONS` `Allow` method if the user so desires to configure it that way.

Something at the ServletContext level tho.
To compliment those users that have Constraint mappings to disable TRACE, so should this default `OPTIONS` method not even show `TRACE` if it's been disabled at the higher level.

Issue #274, opened 2019-12-10
Part.write(String) javadoc does not make sense


janbartel:
Despite the modifications to the javadoc in https://github.com/eclipse-ee4j/servlet-api/issues/172, the text still does not make sense for absolute paths:

>       @param fileName The location into which the uploaded part should
>        be stored. The value may be a file name or a path.  The actual
>        location of the file in the filesystem is relative to {@link
>        javax.servlet.MultipartConfigElement#getLocation()}.  Absolute
>        paths are used as provided and are relative to
>        <code>getLocation()</code>.  Note: that this is a system
>        dependent string and URI notation may not be acceptable on all
>        systems. For portability, this string should be generated with
>        the File or Path APIs.

Absolute paths are used as provided, _but_ yet are relative to the configured location?  Does this mean that if an absolute path is provided, it must begin with the same path as the location?  If it does not, should an IOException be thrown?

markt-asf:
I've just re-read the [EG discussion](https://download.oracle.com/javaee-archive/servlet-spec.java.net/jsr369-experts/2017/03/0558.html) on this. I suggest we use my last proposal in that thread (including the modification Greg suggested) of:
```
@param fileName The location into which the uploaded part should be
                stored. Relative paths are relative to {@link
                javax.servlet.MultipartConfigElement#getLocation()}.
                Absolute paths are used as provided.
                Note: that this is a system dependent string and URI
                notation may not be acceptable on all systems. For
                portability, this string should be generated with
                the File or Path APIs.
```

gregw:
@janbartel I'm happy with that text, so can you make a PR?

joakime:
This seems overlapping with Issue #172

markt-asf:
#172 was the original issue associated with the [EG discussion](https://download.oracle.com/javaee-archive/servlet-spec.java.net/jsr369-experts/2017/03/0558.html) I linked above. In short, the change that was made to "fix" #172 wasn't the one that the EG reached consensus on and included the nonsensical phrase called out in this issue. The same concern was raised in the EG thread but was not addressed. Good to see this fixed.

PR #276, opened 2019-12-11
[274] Corrected javadoc for Part.write(String)

janbartel:
Fix for #274

The javadoc for Part.write(String) contained the contradictory sentence: "Absolute paths are used as provided and are relative to <code>getLocation()</code>."
Changed to the agreed text: "Relative paths are relative to MultipartConfigElement.getLocation(). Absolute paths are used as provided."

### DIFF
--- a/api/src/main/java/javax/servlet/http/Part.java
+++ b/api/src/main/java/javax/servlet/http/Part.java
@@ -74,12 +74,11 @@ public interface Part {
      * implementation to use, for example, file renaming, where possible, rather than copying all of the underlying
      * data, thus gaining a significant performance benefit.
      *
-     * @param fileName The location into which the uploaded part should be stored. The value may be a file name or a
-     *                 path. The actual location of the file in the filesystem is relative to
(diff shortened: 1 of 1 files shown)


joakime:
What is a "relative path" in this?

If I'm running on Microsoft Windows.
And the Multipart location is `D:\web\multipart` and the offered String is `Part.write("/foo.dat");`, then what?

The `/foo.dat` is **not** an absolute path on Windows, as that would result in `D:\foo.dat`.

This needs to be fixed to be extra clear on what "relative path" means.
As that's FileSystem dependent as well.

If the deployment is on a Virtual File System then relative path is again vague.
What if the deployment on a `java.nio.file.FileSystem` that has multiple roots?

gregw:
@joakime a "relative path" is not an "absolute path".  An "absolute path" is described by the javadoc as system dependent, so I by extension, relative paths are also system dependent.    So to determine what is a relative path then APIs like `FileSystems.getDefault().getPath(String)` need to be used to determine that in a system dependent way.

The issue here is that this API should never been created to allow either a filename or a path, as the moment that is allowed things become very system dependent.

PR #277, opened 2019-12-11
Rename package from javax to jakarta

Cousjava:


api/pom.xml
```diff
@@ -34,7 +34,7 @@
 
     <properties>
         <spec.version>4.0</spec.version>
-        <extension.name>javax.servlet</extension.name>
+        <extension.name>jakarta.servlet</extension.name>
         <bundle.symbolicName>jakarta.servlet-api</bundle.symbolicName>
         <bundle.version>4.0.0</bundle.version>
         <window.title>Jakarta Servlet 4.0</window.title>
```
gregw: What is the plan regarding the version number?  I would think we neither wish to keep it the same, nor reset to 1.0.  That really only leaves 4.1 or 5.0 as options?    4.1 would suggest that we have not changed any major functionality, but considering the lack of backwards compatibility it may be best to go with 5.0?

gregw: Ah I see you've gone with 5.0-SNAPSHOT in the parent pom, so you need 5.0 here as well.  Hmmm the bundle version should have been 4.0.2 - perhaps we should set these using properties from the parent pom?

stuartwdouglas: If we do go with 5.0 as the version number then it should be changed everywhere, I don't think it makes sense to have 5.0 as the artifact version but 4 as the bundle version.

stuartwdouglas:
Other than the version issue I think this looks good. IMHO we should try and merge this ASAP, and any fixes can come in later.

PR #279, opened 2019-12-23
Fix build by adding missing imports

stuartwdouglas:
In the rename this class seems to have lost some imports,
so the build currently fails.


stuartwdouglas:
There was an 'import javax.servlet.*;' that was removed.

PR #280, opened 2019-12-23
Use 5.0 version

stuartwdouglas:

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -27,7 +27,7 @@
 
     <groupId>jakarta.servlet</groupId>
     <artifactId>jakarta.servlet-api</artifactId>
-    <version>4.0.4-SNAPSHOT</version>
+    <version>5.0-SNAPSHOT</version>
     
     <name>Jakarta Servlet</name>
     <url>https://projects.eclipse.org/projects/ee4j.servlet</url>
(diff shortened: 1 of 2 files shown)


api/pom.xml
```diff
@@ -27,7 +27,7 @@
 
     <groupId>jakarta.servlet</groupId>
     <artifactId>jakarta.servlet-api</artifactId>
-    <version>4.0.4-SNAPSHOT</version>
+    <version>5.0-SNAPSHOT</version>
```
bshannon: The API jar file version should be 5.0.0-SNAPSHOT.
The spec document version should be just 5.0-SNAPSHOT (as it is below).

stuartwdouglas: @bshannon can all the maven artifact versions be 5.0.0? I think I misunderstood you originally and thought you meant that the spec artifacts must be 5.0 while the API 5.0.0, but on re-reading did you just mean the spec.version property rather than the actual maven artifact version?

bshannon: All technical artifacts (jar files) use dot-dot version numbers (e.g., 5.0.0).

All specification documents use single dot version numbers (e.g., 5.0).

The version number of the API jar file must track the version number of the specification.

The implementation jar files may or may not track the version number of the specification.

More information here:
https://wiki.eclipse.org/JakartaEE_Maven_Versioning_Rules

PR #281, opened 2019-12-23
Add Github actions CI

stuartwdouglas:
This will test master and pull requests.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,17 @@
+name: Java CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml


.github/workflows/maven.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
```
gregw: Since 5.0 will be java >= 11 should this be java 11?

stuartwdouglas: Our build still requires JDK8. Do you want me to update the build to JDK11?

gregw: I'm OK with doing one thing at a time:  ie get CI going and then change build to JDK11.... Oh you've changed to 11... that's fine too

stuartwdouglas: I changed it to 11 then realised it failed. I might change it back to 8, then change to 11 as a different PR

stuartwdouglas: I have changed it back to 8 and the build passed: https://github.com/stuartwdouglas/servlet-api/commit/1b9375592c7c122d36932a02caf69db16ee91fd6/checks?check_suite_id=371172746

Once it is actually merged I think it will show up as a proper status check.

PR #282, opened 2019-12-23
Change version to 5.0.0-SNAPSHOT

stuartwdouglas:


gregw:
If we want to change to 5.0.0-SNAPSHOT, then I think there are more places this needs to be changed.  In the child poms at the very least.

stuartwdouglas:
This is just the API ,  it was based on Bill's comment at https://github.com/eclipse-ee4j/servlet-api/pull/280#discussion_r360747097 that the API jar should be 5.0.0 while the others should be 5.0. Another inconsistency I noticed is that the API does not actually use the Servlet parent but the eclipse parent, but as there is nothing really important in the Servlet parent I think we may want to keep it like this.

stuartwdouglas:
I would also rather version everything as 5.0.0, I only went with 5.0 with the initial pull request to be consistent.

bshannon:
> This is strange to me. I don't see why the spec document can't also be versions with dot releases, but let's go with the flow for now.

Single dots, yes, double dots, no.

We don't want someone claiming that they're compatible with Servlet 5.0.0 and someone else claiming that they're compatible with Servlet 5.0.1 and thus "better" or "more compatible".  A maintenance release of a spec document is only for doing simple things like fixing typos or other changes that don't actually change the requirements.  Think of it as fixing errata.  It's like a new printing of a book.  It's the same book, with the same story, but some trivial things have been fixed.  Because of that, we don't want to distinguish it with a new version number that implies new or updated requirements.  Instead we use a "rev level",  For compatibility purposes, it shouldn't matter which rev level of the spec you're looking at.

PR #284, opened 2020-01-06
Legal files fixed for multi module project structure

senivam:
jakarta-servlet-api.jar shall contain:
 / -
    META-INF /
        LICENSE.md
        NOTICE.md

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 1997, 2018 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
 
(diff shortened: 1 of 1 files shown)


api/pom.xml
```diff
@@ -45,6 +45,7 @@
         <findbugs.exclude>exclude.xml</findbugs.exclude>
         <findbugs.threshold>Low</findbugs.threshold>
         <copyright-plugin.version>1.49</copyright-plugin.version>
+	<legal.source.folder>${maven.multiModuleProjectDirectory}</legal.source.folder>
```
bshannon: Who sets maven.multiModuleProjectDirectory?

Based on a little googling, it doesn't appear that this is a built in Maven property.
It looks like it's something that every user needs to set when invoking "mvn".
Is that true?  If so, I recommend using this plugin instead:

                <groupId>org.commonjava.maven.plugins</groupId>
                <artifactId>directory-maven-plugin</artifactId>

PR #285, opened 2020-01-07
Using directory-maven-plugin to initialize legal.source.folder property

senivam:
improving #284 by using plugin to initialize property instead of internal maven property 


senivam:
@bshannon - required change implemented, could you please take a look?

senivam:
yes, you are right, the target for plugin shall be adjusted in order to work from the lower lever directory. However in order to work, the api's parent shall be adjusted as well. So, I've done required modifications to allow property initialization from any folder.

PR #289, opened 2020-01-20
Fix #288 add copyright checks

### LINKED ISSUE #288: Automate checkstyle and copyright
The build / CI / commit process should automatically apply checkstyler rules and verify copyright year has been updated.

markt-asf:
The plug-in first states the dates in most files needs to be updated to 2019. Do that and it then says it needs to be 2020 (because they have changed again). Given that these files are going to be released/published in 2020, I think a 2020 date is fine.

For the actual text, the plug-in complains if the "and others" part is missing. This only affected a handful of files. Looking at them, they had all been modified (or at least moved) since the initial import so the "and others" text seems justified to me.

### DIFF
--- a/api/etc/config/copyright-exclude
+++ b/api/etc/config/copyright-exclude
@@ -1,0 +1,4 @@
+.png
+speclicense.html
+/etc/config/copyright-exclude
+/etc/config/copyright.txt
--- a/api/etc/config/copyright.txt
+++ b/api/etc/config/copyright.txt
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) YYYY Oracle and/or its affiliates and others.
+ * All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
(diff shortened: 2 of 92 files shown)


markt-asf:
As per the Eclipse project handbook, it is the "1997, 2020" form that should be used. I'll go through and make them consistent and update the PR.

arjantijms:
@gregw @markt-asf 
> As per the Eclipse project handbook, it is the "1997, 2020" form that should be used. I'll go through and make them consistent and update the PR.

Very good that this came up, as I didn't realise this before. I'll check this for Faces, Authentication etc as well.

bshannon:
> While I know many projects (mine included) do update the copyright year on all files every year, I have seen @bshannon say that is not the policy for Jakarta projects and that we should only update on files that have been changed..... Oh but I guess we have changed every file because of the name.... never mind!

This is the advice I've gotten from our lawyers.  I'm quite sure that if you ask other lawyers you'll get different advice.  The only "good" reason to do the bulk update was that people will forget to do the update when they should.  A reason that the bulk update is "bad" is that it clutters up the history of files with lots of gratuitous commits.  Some people care more about keep a clean history than others.

> Ah but we do have a mix of styles: `1997-2020` vs `1997, 2020`. I much prefer the former if the plugin can handle it. Eitherway we should be consistent.

The plugin handles it, and I prefer it too, but the Eclipse (and Oracle) style is the latter.

markt-asf:
I've pushed an updated PR that fixes the 2 files I left in the wrong format.

PR #290, opened 2020-01-27
Remove unnecessary imports

pzygielo:


pzygielo:
@markt-asf thanks for checking.

I don't believe javadoc ever had issues with references in the same package, in the same source tree. But it'd be extremely funny should it be true.

gregw:
Note that the [eclipse style file](https://raw.githubusercontent.com/eclipse-ee4j/ee4j/master/codestyle/ee4j_intellij_formatting.xml) doesn't include an ordering for imports, so if applied to the intellij default you get the order javax, java, other.  This changes 26 files.

pzygielo:
This PR aims at fixing only one CS violation, which is [RedundantImport](https://checkstyle.sourceforge.io/config_imports.html#RedundantImport):
https://github.com/eclipse-ee4j/servlet-api/runs/413209896#step:4:1792

bshannon:
> Note that the [eclipse style file](https://raw.githubusercontent.com/eclipse-ee4j/ee4j/master/codestyle/ee4j_intellij_formatting.xml) doesn't include an ordering for imports, so if applied to the intellij default you get the order javax, java, other. This changes 26 files.

Is this a typo?  Does IntelliJ really do javax, java, other?  Above you mentioned java, javax, other,
which seems like the right order.

gregw:
On my IDE (2019.3.1):
+ default style gives javax, java, other as the import order with no blank lines between
+ EE4J as imported by me gives javax, java, blank line, other.... which is strange because the https://github.com/eclipse-ee4j/ee4j/blob/master/codestyle/ee4j_intellij_formatting.xml file does not include import ordering?

So perhaps this is strangeness in how my IDE is configured?   But either way, I think the solution is to be explicit in the XML file about what the ordering should be (java, javax, other) , so maybe a PR on EE4J is needed to add that? 

And more importantly.... shouldn't that ordering now include jakarta:  (java, javax, jakarta, other)?

gregw:
Correction:
 + My IDEA default is: blank, javax, java, others, blank, statics
 + EE4J on my IDEA is: others, blank, javax, java, blank, statics

I'd like to propose that EE4J is: java, javax, jakarta, others, blank, statics and that we should manually configure our IDEs to that until the EE4J style is updated. 

Sorry @pzygielo to hijack what probably felt like a simple PR :(

pzygielo:
> Sorry @pzygielo to hijack what probably felt like a simple PR :(

No worries. Formatting with preferences from ee4j/codestyle does not remove those imports so even with #298 it still is simple and valid PR :wink:

PR #291, opened 2020-01-27
Remove overriding method

pzygielo:


markt-asf:
I agree that this look unnecessary. Again, I'm happy to approve this but I'm going to leave it a little while to give others a chance to comment if they wish.

stuartwdouglas:
I also agree that it is unnecessary, but I think it will also require an update of the TCK signature tests?

bshannon:
> I also agree that it is unnecessary, but I think it will also require an update of the TCK signature tests?

Yes, it probably will.

gregw:
Whilst we have not really finished the 5.0 vs 5.1 discussion yet, I have marked this as a 5.1 thing, as I think the current intent is to make 5.0 identical to 4.0 in everything but name.  This kind of change is potentially something that could trip up a binary conversion tool.

I think there is still some chance that we will decide to allow minor changes in 5.0, but until we do let's mark such PRs as 5.1.   If we don't soon decide to allow minor changes in 5.0, then we need to create branches for 5.0 and 5.1 to target these kind of PRs at.

markt-asf:
Removing an overriding method is a binary compatible change. I have performed a test locally to confirm this.

It will require an update to the signature tests in the TCK but as we are going to have to update those anyway firstly for the `javax` -> `jakarta` package name and secondly to provide Java 8 and java 11 versions of the signature files (caused by the changes to Object based on what I have seen so for for EL) I don't view that as a reason not to make this change.

Any automated Jakarta EE 8 to Jakarta EE 9 conversion tool will be unaffected by removing this method because the change is a binary compatible one.

I am strongly in favour of including this change in 5.0.0.

arjantijms:
I agree with Mark. Binary compatible changes such as adding generics or removing needless overrides are a good candidate for this version.

For the ones that remember, we had planned and done these initially before Jakarta EE 8 even, but didn't do them there because of the strict TCK requirements.

Issue #292, opened 2020-01-28
Import Specification document for Jakarta Servlet


gregw:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21457

PR #293, opened 2020-01-28
Import Jakarta Servlet Specification document #292

gregw:
For issue #292:
 + Fixed pom hierarchy (should parent be 5.0.0 or 5.0?)
 + Concatenated spec document with boilerplate in servlet-spec.adoc
 + Updated copyright to 2020
 + Updated references to 4.0 to 5.0


gregw:
Note that this document does not format very well: example formatting is wrong, images are poor resolution, etc.  So a lot of work is required, but I thought it best to capture more or less as contributed.

arjantijms:
> Note that this document does not format very well: example formatting is wrong, images are poor resolution, etc. So a lot of work is required, but I thought it best to capture more or less as contributed.

That holds for many if not all documents indeed. It's quite a bit of work to rectify / clean all of this. See also the discussion at https://www.eclipse.org/lists/jakartaee-platform-dev/msg01409.html

markt-asf:
Agreed on the view that we should commit this largely as provided and then fix it. +1 to apply this PR.

Only yesterday I did a three-way diff between the spec files (minus the main spec doc) for EL, WebSocket and JSP and cleaned up various inconsistencies and shared minor improvements across all of them. I'd be happy to provided a similar PR for Servlet.

gregw:
oops sorry, I should not have merged until @stuartwdouglas approved (unless @markt-asf has been voted in already).

pom.xml
```diff
@@ -26,10 +26,10 @@
         <version>1.0.5</version>
     </parent>
 
-    <groupId>org.eclipse.ee4j.servlet</groupId>
+    <groupId>jakarta.servlet</groupId>
     <artifactId>servlet-parent</artifactId>
     <packaging>pom</packaging>
-    <version>5.0-SNAPSHOT</version>
```
stuartwdouglas: I think this is incorrect, I believe the document should not support micro versions, as per a comment from a previous PR.

stuartwdouglas:
I don't know if we can have 5.0.0 as the pom version based on the comment at https://github.com/eclipse-ee4j/servlet-api/pull/282.

Otherwise looks good to me. @markt-asf if you are happy to do the initial clean up that is fine with me, we just need to be careful we don't have multiple people working on it at once as that will just cause massive conflicts.

markt-asf:
Just to be clear, the clean-up I was volunteering for covers the spec files *excluding* the document that contains the bulk of the text.
I'm happy to help out with the main doc too but I have other things ahead of that on my TODO list - mainly the entire JSP spec doc - so I won-t be in a position to help with the main servlet doc for a while.

bshannon:
> I don't know if we can have 5.0.0 as the pom version based on the comment at #282.

The spec document version should be 5.0, or 5.0-SNAPSHOT while you're working on it.

gregw:
So for clarity... after this PR we have:
 + parent pom 5.0.0
 + API pom 5.0.0
 + Spec pom 5.0
 
So the spec is 5.0 not 5.0.0.  This issue is should the parent pom have micro versions or not?   There is not a lot of content in the parent pom, but it's sibling files (eg `CONTRIBUTING.md` and `README.md`) are likely to go through lots of revisions  hence I thought the micro version was appropriate... but these are not released artefacts, so it makes no difference and thus the parent should probably just be 5.0.

I think I updated most 4.0 to 5.0 in the document, except for some examples referencing schemas/dtd.

So @markt-asf in your new role as leader, is this the kind of trivial fix that you had in mind for direct commits to master :)   I should I do another PR?

stuartwdouglas:
Oops, looks like I miss-read the PR, for some reason I thought it was the spec you changed but it was the parent. 

Don't we have to release the parent in order to make maven central happy? I don't think it is good to release an artifact with a non-resolvable parent, so I think the current state is fine.

markt-asf:
Not being at all familiar with Maven, I'm not entirely sure what role the top level POM is performing. I think it is there to provide a short-cut to build the api, impl (where projects have one - we don't) and spec together rather than to act as a parent to the api, impl and spec POMs. I think it will be simpler to have those POMs point to the Jakarta EE parent (which by the way is now a 1.0.6 and I'm reasonable sure we need to update to it to be able to release milestones and/or RCs).

Wearing my committer hat, I'd be fine with this being a direct commit. I don't know about other committers. They may be less comfortable. If we enable option for direct commit then I expect there will be a period of us figuring out what, as a community, we are happy with it being used for.

bshannon:
> So for clarity... after this PR we have:
> 
>     * parent pom 5.0.0
> 
>     * API pom 5.0.0
> 
>     * Spec pom 5.0
> 
> 
> So the spec is 5.0 not 5.0.0.

Correct.

> This issue is should the parent pom have micro versions or not? 

Yes, it should.

arjantijms:
Hi,

On Thursday, January 30, 2020, Stuart Douglas <notifications@github.com>
wrote:

> Oops, looks like I miss-read the PR, for some reason I thought it was the
> spec you changed but it was the parent.
>
> Don't we have to release the parent in order to make maven central happy?
> I don't think it is good to release an artifact with a non-resolvable
> parent, so I think the current state is fine.
>
A parent always needs to be released. Maven Central won’t complain
initially, but clients do.

We once bodged a JSF release because of this :(

That’s why in most cases the independent spec and api modules are better
off not referencing the parent IMHO.

Kind regards,
Arjan

PR #294, opened 2020-01-28
Initial merge of spec document

### LINKED ISSUE #292: Import Specification document for Jakarta Servlet

stuartwdouglas:
Fixes #292

This is an initial import of the spec file largely as it was provided to us but modified slightly to fit into the existing build (file was just renamed and included into the existing doc).

Some questions we need to answer before we merge though:

- Do we need Servlet.txt? I don't think so but I included it
- Do we try and fix it up before the merge, or just merge as provided and then fix? There is lots of potential for conflict here if multiple people are involved.


gregw:
This is a duplicate of #293.   I had hoped to avoid that by assigning issue to myself - oops

gregw:
I don't think we need Servlet.txt
I like your way of including the body better than my concatenation
I think we need to fix the poms to work

I'll change mine to use your include method

stuartwdouglas:
I had already done it locally when you filed the issue, I was not expecting you to be so quick :-)

PR #297, opened 2020-01-30
Use spotless to check formatting

pzygielo:


gregw:
What are the relative merits of spotless vs checkstyle?     
Has spotless been picked to follow the rest of the EE herd or by some other criteria?

pom.xml
```diff
@@ -61,4 +61,31 @@
...
+                            <id>spotless-check</id>
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                    <configuration>
+                        <java>
+                            <eclipse>
+                                <file>https://raw.githubusercontent.com/eclipse-ee4j/ee4j/master/codestyle/ee4j-eclipse-formatting.xml</file>
```
gregw: Can this reference be to a release tag rather than master so that we have a durable build.

pzygielo:
> What are the relative merits of spotless vs checkstyle?

It supports eclipse formatter preferences presented at [eclipse-ee4j/ee4j:codestyle](https://github.com/eclipse-ee4j/ee4j/tree/master/codestyle).

> Has spotless been picked to follow the rest of the EE herd 

I don't know. Can this project decide for itself?

> or by some other criteria?

By criterion of using eclipse formatter that can be included in automated build instead of clicking in IDE.

gregw:
> It supports eclipse formatter preferences presented at [eclipse-ee4j/ee4j:codestyle](https://github.com/eclipse-ee4j/ee4j/tree/master/codestyle).
> 
> I don't know. Can this project decide for itself?
> 
> By criterion of using eclipse formatter that can be included in automated build instead of clicking in IDE.

I'm sure this project can decide for itself.  I've got no preference either way and was just curious to why you selected spotless.  The ability to import an eclipse formatting file sounds like a good one to me.

However, taking your reformatted branch and applying the IDEA EE4J style results in lots of the formatting being reverted.  I think more work is needed on the IDEA formatting style to bring it into line with the eclipse one.

pom.xml
```diff
@@ -61,4 +61,31 @@
...
+                            <id>spotless-check</id>
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                    <configuration>
+                        <java>
+                            <eclipse>
+                                <file>https://raw.githubusercontent.com/eclipse-ee4j/ee4j/${project.parent.parent.version}/codestyle/ee4j-eclipse-formatting.xml</file>
```
pzygielo: It was the worst I could do. In local parent refer to parent.parent because it works in child :cry:
I'll move configuration down to api.

stuartwdouglas:
We work around the Intellij formatter problems in Quarkus by using the Eclipse formatter plugin for Intellij, so both IDE's use the same format under the hood. It was the only way we could find to get a 100% consistent format, as there are subtle differences between the formatter implementations.

We then also run the formatter as part of a normal build, while CI checks the formatting rather than performing. It is a bit different to how developers are used to working but it has actually worked quite well.

pzygielo:
> We then also run the formatter as part of a normal build, while CI checks the formatting rather than performing. 

Reworded commit message to not suggest that spotless:apply is run automatically.

markt-asf:
When I did this previously under #199, I took the Eclipse config file that was on the wiki (and is now [in GitHub](https://github.com/eclipse-ee4j/ee4j/blob/master/codestyle/ee4j-eclipse-formatting.xml)), configured it in Eclipse (I use Eclipse as my IDE) ran it over the code the generated a PR from the diif. When I do that now, I get very different results (less indentation).

pzygielo:
May I ask to move it forward, please? (In any direction.)

stuartwdouglas:
So I have an alternative approach at https://github.com/eclipse-ee4j/servlet-api/pull/312 

The end result is basically the same, but this one will automatically fix problems when you build (but not on CI). 

IMHO this approach is a lot more user friendly, as new contributors don't really have to deal with formatting errors, but I am happy to go with whatever the consensus is.

Whichever approach we go with we need a local copy of the formatting file, otherwise changes to this file can break our build and make old commits unbuildable.

pzygielo:
@stuartwdouglas thanks for checking

> So I have an alternative approach at #312

:+1:

PR #298, opened 2020-01-30
Apply spotless

pzygielo:


gregw:
This PR contains mostly formatting changes.  I think now is a good time to get the formatting right, but I'd like to know why the formatting applied here is so different to that applied in #199?   The last thing we need is a battle between two formatting styles. @pzygielo  Can you explain how the formatting was applied here? and @markt-asf can you also confirm exactly how you did the formatting in #199 

I do note that the style for intellij from https://github.com/eclipse-ee4j/ee4j suffers from the issue that it only includes settings that differ from the default, so that the resulting style can depend on how the IDE was configured before it was applied.

gregw:
Also, shouldn't any formatting changes be part of #297 rather than a PR for spot bugs?

pzygielo:
> This _draft_ PR contains mostly formatting changes.

It's based on #297, with two additional commits. (I planned to rebase that pair in case #297 was accepted.)

> but I'd like to know why the formatting applied here is so different to that applied in #199?

I do not know that.

> The last thing we need is a battle between two formatting styles.

Couldn't agree more with above.

> Can you explain how the formatting was applied here?

No human intervention. Only as second to last commit [almost] says:
```
mvn spotbugs:apply -am -pl api
```
configured with https://github.com/eclipse-ee4j/servlet-api/blob/04190ca22e057219bf032c9e2f952079f2b4611d/pom.xml#L81-L83


> Also, shouldn't any formatting changes be part of #297 rather than a PR for spot bugs?

I'd prefer to have them split. But if you consider this fuller chain of commits as better solution I can mark this PR as ready for review and close the earlier.

gregw:
> I'd prefer to have them split. But if you consider this fuller chain of commits as better solution I can mark this PR as ready for review and close the earlier.

I think having them split is also good.  But I would have thought the spotless configuration and formatting changes should all have been in one PR (#297) and the spot bugs configuration should be in this PR.

If you can commit the reformat to #297 then we can have the discussion just there about the format.

pzygielo:
> and the spot bugs configuration should be in this PR.

Nice catch. My repeated mistake - it's all about spotless now only.
Spotbugs can be next :smile: (Given there are no findings anyway AFAIK.)
Will fix commit message and push to #297.

pzygielo:
Formatting commits pushed to #297.

Issue #295, opened 2020-01-28
Update CONTRIBUTING.md with guidelines for commits, PRs, reviews and merges


gregw:
The `CONTRIBUTING.md` has the following:
```
Before your contribution can be accepted by the project team contributors must
electronically sign the Eclipse Contributor Agreement (ECA).

* http://www.eclipse.org/legal/ECA.php

Commits that are provided by non-committers must have a Signed-off-by field in
the footer indicating that the author is aware of the terms by which the
contribution has been provided to the project. The non-committer must
additionally have an Eclipse Foundation account and must have a signed Eclipse
Contributor Agreement (ECA) on file.
```

I think we need to consider having a lot more in there:

Code
====
 + code should use https://github.com/eclipse-ee4j/ee4j code style
 + Do we want to run check style also?

Pull Requests
==========
 + Should PRs from committers use branches within the repo or come from personal forks?  

Merging
======
 + CI must pass
 + Must be reviewed by at least 1 project leader?
 + Should we have a minimum age so that committers in all timezones PRs have a chance to review?
 + Should any committer be able to call for a formal `-1`, `0`, `+1` style vote on a PRs (as is described in the [EDP](https://www.eclipse.org/projects/dev_process/#4_7_Committers_and_Contributors))?  
 + With votes, who is it that may adjudicate if a `-1` is an `Invalid Veto` as per the process? 

I think we need to discuss what guidelines we want, formulate them into a PR on  `CONTRIBUTING.md` and then have an EDP style vote on it to make it official.

gregw:
My thoughts are as follows:
 + checkstyle must pass
 + CI must pass
 + PRs can come from branches in any repo.  Committers must cleanup branches in the main repo.
 + Any committer can call for a formal vote in a PR comment, which will be take at 1 week and be adjudicated by the project lead(s).
 + Non voted on PRs can be merged after being non-draft for 1 week and having 2 reviews from committers from a different organisation(s) to the author(s).  
 + Project leads may bypass any of these criteria, but must document their reasons for doing so.

markt-asf:
I agree with your views on CheckStyle, CI and PR branches.
I'd say that any committer can override those criteria as long as they document the reason. Implicit in that is that if the consensus is that the reason isn't valid, then the commit may need to be reverted. (Other options include an additional correcting commit, leave as-is but agree to use a PR for future similar changes etc.)
[...]

PR #299, opened 2020-01-30
Issue #295 Contributing Guidelines

gregw:
#295

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,12 +40,13 @@ https://www.eclipse.org/projects/handbook/#resources-commit
 
  * Code should adhere to the style defined at https://github.com/eclipse-ee4j/ee4j.
  * Contributions should be made via github pull requests at https://github.com/eclipse-ee4j/servlet-api/pulls 
- * Pull Requests should have reviews requested from several project committers from different organisations than the author(s). 
- * Pull Request by committers that receive no review responses after a reasonable time, considering time zones and working days, can be interpreted as approved.
+ * Authors of Pull Requests should seek review from committers outside of their own organisation and perspective.
+ * Committers and Contributors may review any Pull Requests or Commits, even if not requested or assigned.
+ * Pull Request by Committers that receive no review responses after a reasonable time, considering time zones and working days, can be interpreted as approved.
  * Pull Requests should pass continuous integration tests prior to merging.
  * Pull Requests for significant contributions, new features or changed behaviour should be open for comment and review in a non draft state for at least 1 week prior to merging.
- * For disputed Pull Requests, any committer may request a formal consensus vote (ie. +1, 0, -1). Any such vote will be scheduled, conducted and adjudicated by the Project Leader(s).
- * If a disputed Pull Request is merged prior to a consensus being established, then it shall be considered reverted and consensus must be sought for it to remain rather than for it's reversion.
+ * In exceptional cases when a consensus on a Pull Request or Commit cannot be established through normal review and discussion, then any committer may request a formal consensus vote (ie. +1, 0, -1). Any such vote will be scheduled, conducted and adjudicated by the Project Leader(s).
+ * If the consensus for a merged Pull Request or commit is disputed, then it shall be considered reverted and consensus must be sought for it to remain rather than for it's reversion.
  * In trivial or exceptional circumstances, these guidelines may be bypassed if the reasons for doing so are documented.
 
 ## Contact
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,18 @@ Contributor Agreement (ECA) on file.
 For more information, please see the Eclipse Committer Handbook:
 https://www.eclipse.org/projects/handbook/#resources-commit
 
+## Project Contribution Guidelines
+
+ * Code should adhere to the style defined at https://github.com/eclipse-ee4j/ee4j.
+ * Contributions should be made via github pull requests at https://github.com/eclipse-ee4j/servlet-api/pulls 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -36,6 +36,18 @@ Contributor Agreement (ECA) on file.
 For more information, please see the Eclipse Committer Handbook:
 https://www.eclipse.org/projects/handbook/#resources-commit
 
+## Project Contribution Guidelines
+
+ * Code should adhere to the style defined at https://github.com/eclipse-ee4j/ee4j.
+ * Contributions should be made via github pull requests at https://github.com/eclipse-ee4j/servlet-api/pulls 
+ * Pull Requests should have reviews requested from several project committers from different organisations than the author(s).
```
markt-asf: Without knowing the availability and current commitments of other committers I'm not sure how I'd identify appropriate committers to assign to the review. I'm also concerned that it may encourage other committers not to review the PR because someone else has been assigned. Finally, I think the "different organisations" part could be read as "we don't trust committers with the same employer not to collude to the advantage of their employer". I'm wondering whether the following point is sufficient to make this one unnecessary? Or maybe explicitly state that reviewing PRs is the responsibility of all committers but we also welcome constructive reviews from non-committers?

CONTRIBUTING.md
```diff
@@ -36,6 +36,18 @@ Contributor Agreement (ECA) on file.
...
 
+## Project Contribution Guidelines
+
+ * Code should adhere to the style defined at https://github.com/eclipse-ee4j/ee4j.
+ * Contributions should be made via github pull requests at https://github.com/eclipse-ee4j/servlet-api/pulls 
+ * Pull Requests should have reviews requested from several project committers from different organisations than the author(s). 
+ * Pull Request by committers that receive no review responses after a reasonable time, considering time zones and working days, can be interpreted as approved.
+ * Pull Requests should pass continuous integration tests prior to merging.
+ * Pull Requests for significant contributions, new features or changed behaviour should be open for comment and review in a non draft state for at least 1 week prior to merging.
+ * For disputed Pull Requests, any committer may request a formal consensus vote (ie. +1, 0, -1). Any such vote will be scheduled, conducted and adjudicated by the Project Leader(s).
```
markt-asf: May add something like "... attempts should first be made to resolve issues through discussion. If consensus can't be reached after a reasonable period of time then...". Of course, that raises the question "How long is reasonable?". I'd probably leave that vague as I think it will vary from case to case.

CONTRIBUTING.md
```diff
@@ -36,6 +36,18 @@ Contributor Agreement (ECA) on file.
...
+ * If a disputed Pull Request is merged prior to a consensus being established, then it shall be considered reverted and consensus must be sought for it to remain rather than for it's reversion.
```
markt-asf: I like this.

CONTRIBUTING.md
gregw: I can see how that reads negatively, but it is really meant as a positive guideline to seek review from outside ones own bubble.   It wasn't really meant to be about the mechanics of requesting a review.
How about breaking it into two points:
+ Authors of Pull Requests should seek review from committers outside of their own organisation and perspective.
+ Committers and Contributors may review any Pull Requests, even if not requested or assigned.

CONTRIBUTING.md
gregw: I put in the Scheduled by project leader bit so they could allow a reasonable time for discussion in case somebody was too quick to call a vote.   But I agree the guidelines should indicate that formal votes are exceptional. How about:
+ In exceptional cases when a consensus on a Pull Request cannot be established through normal review and discussion, then any committer may request a formal consensus vote (ie. +1, 0, -1). Any such vote will be scheduled, conducted and adjudicated by the Project Leader(s).

CONTRIBUTING.md
gregw: yep - we don't want to encourage a rush to merge contentious changes.  Actually I think it should be reworded to:
* If the consensus for a committed/merged change is disputed, then it shall be considered reverted and consensus must be sought for it to remain rather than for it's reversion.

CONTRIBUTING.md
```diff
@@ -36,6 +36,19 @@ Contributor Agreement (ECA) on file.
 For more information, please see the Eclipse Committer Handbook:
 https://www.eclipse.org/projects/handbook/#resources-commit
 
+## Project Contribution Guidelines
+
+ * Code should adhere to the style defined at https://github.com/eclipse-ee4j/ee4j.
+ * Contributions should be made via github pull requests at https://github.com/eclipse-ee4j/servlet-api/pulls 
+ * Authors of Pull Requests should seek review from committers outside of their own organisation and perspective.
+ * Committers and Contributors may review any Pull Requests or Commits, even if not requested or assigned.
+ * Pull Request by Committers that receive no review responses after a reasonable time, considering time zones and working days, can be interpreted as approved.
```
stuartwdouglas: How will this work in practice, as commiters can't merge their own commits?

gregw: I assume that if this is adopted we want it common place, then we'd have to ask to have the requirement for a review to be removed from github.
Note that I think committers are able to merge locally and push to master or any branch, so they can always bypass the github check if they really want to.

Personally I don't mind having to seek review, so I'm happy not to have this.  But if others want it, I think the rest of the guidelines provide enough of a safety net that problems that arise could be handled.

stuartwdouglas:
I think this looks good for now, we can always update it later if we want changes.

CONTRIBUTING.md
markt-asf: That works for me

PR #306, opened 2020-02-13
More formatting

stuartwdouglas:
- Fix all sections so section numbers are correct
- Fix all bulleted and numbered lists
- Fix all code blocks
- Fix all tables
- Move all footnotes to asciidoc footnote

Note that Appendix A and Section 1.4 were skipped, as the intention is to remove them.


stuartwdouglas:
@markt-asf @gregw this fixes up all the section headers, so the numbers now line up with the original spec, as well as a heap of other formatting fixes.

Once this is in I am going to try and do at least one chapter a day to get the formatting sorted out.

I have only been looking at formatting so far, I have not made any attempt to update Java -> Jakarta.

stuartwdouglas:
I have now fixed the formatting for every chapter, except for Appendix A and 14.4, both of which are planned for deletion.

stuartwdouglas:
@markt-asf @gregw could I get a review on this? I don't really want to spend any more time on the spec document until this has been merged.

arjantijms:
Thanks for all the work, it's a lot! (I did similar work for Authentication and Authorization, it's quite boring and tiring)

PR #310, opened 2020-04-07
Add JPMS descriptor

Thihup:
Signed-off-by: Thiago Henrique Hüpner <thihup@gmail.com>

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -278,6 +278,9 @@
                         </goals>
                         <configuration>
                             <source>1.8</source>
+                            <sourceFileExcludes>
+                                <sourceFileExclude>module-info.java</sourceFileExclude>
+                            </sourceFileExcludes>
                             <additionalJOption>-Xdoclint:none</additionalJOption>
                             <description>Jakarta Servlet API documentation</description>
                             <doctitle>Jakarta Servlet API documentation</doctitle>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -278,11 +278,6 @@
                         </goals>
                         <configuration>
                             <source>1.8</source>
-                            <release>8</release>
-                            <sourceFileExcludes>
-                                <sourceFileExclude>module-info.java</sourceFileExclude>
-                            </sourceFileExcludes>
-                            <notimestamp>true</notimestamp>
                             <additionalJOption>-Xdoclint:none</additionalJOption>
                             <description>Jakarta Servlet API documentation</description>
                             <doctitle>Jakarta Servlet API documentation</doctitle>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -146,10 +146,25 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
-                    <compilerArgument>-Xlint:all</compilerArgument>
+                    <release>9</release>
+                    <compilerArgs>
+                        <arg>-Xlint:all</arg>
+                    </compilerArgs>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>base-compile</id>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                        <configuration>
+                            <release>8</release>
+                            <excludes>
+                                <exclude>module-info.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
             
             <!-- Checks copyright / license headers -->        
@@ -223,9 +238,6 @@
                 <configuration>
                     <archive>
                         <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-                        <manifestEntries>
-                            <Automatic-Module-Name>java.servlet</Automatic-Module-Name>
-                        </manifestEntries>
                     </archive>
                     <excludes>
                         <exclude>**/*.java</exclude>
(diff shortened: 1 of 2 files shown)


gregw:
@sbordet can you please review this

api/pom.xml
```diff
@@ -257,6 +269,11 @@
                         </goals>
                         <configuration>
                             <source>1.8</source>
+                            <release>8</release>
+                            <sourceFileExcludes>
+                                <sourceFileExclude>module-info.java</sourceFileExclude>
+                            </sourceFileExcludes>
+                            <notimestamp>true</notimestamp>
```
joakime, line 276: This configuration on the javadoc-plugin seems out of scope for this PR.

api/pom.xml
```diff
@@ -135,12 +135,27 @@
...
+                    </compilerArgs>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>base-compile</id>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                        <configuration>
+                            <release>8</release>
```
gregw, line 152: I don't think we need or should build with java8.   That will only result in different builds for the same version of the api.

api/src/main/java/module-info.java
```diff
@@ -0,0 +1,6 @@
+module jakarta.servlet {
+    exports jakarta.servlet;
+    exports jakarta.servlet.annotation;
+    exports jakarta.servlet.descriptor;
+    exports jakarta.servlet.http;
```
gregw, line 5: Jetty has been adding JPMS to our re-bundling of 4.0 API and we have included an export for jakarta.servlet.resources which includes all the XSDs and DTDs.    I believe tomcat also bundles the resources the same way.

I don't think we really want to have another JPMS module involved that has the XSDs and DTDs... if so should this module depend on that?

api/pom.xml
```diff
@@ -214,9 +229,6 @@
                 <configuration>
                     <archive>
                         <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
```
gregw, line 231: Probably out of scope for this PR, but we really should also add the OSGi headers in the manifest.

api/pom.xml
joakime: I thought all Jakarta EE9 spec artifacts were supposed to be Java 8 bytecode.

api/pom.xml
joakime: The `/api/` build here already adds OSGi manifest headers.

The result of `mvn clean install` on `master` branch today.

```
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Created-By: Apache Maven Bundle Plugin
Built-By: joakim
Build-Jdk: 11.0.6
Automatic-Module-Name: java.servlet
Bnd-LastModified: 1586271343407
Bundle-Description: Jakarta Servlet 5.0
Bundle-DocURL: https://www.eclipse.org
Bundle-License: http://www.eclipse.org/legal/epl-2.0, https://www.gnu.or
 g/software/classpath/license.html
Bundle-ManifestVersion: 2
Bundle-Name: Jakarta Servlet
Bundle-SymbolicName: jakarta.servlet-api
Bundle-Vendor: Eclipse Foundation
Bundle-Version: 5.0.0
Export-Package: jakarta.servlet;uses:="jakarta.servlet.annotation,jakart
 a.servlet.descriptor";version="5.0.0",jakarta.servlet.annotation;uses:=
 "jakarta.servlet";version="5.0.0",jakarta.servlet.descriptor;version="5
 .0.0",jakarta.servlet.http;uses:="jakarta.servlet";version="5.0.0"
Extension-Name: jakarta.servlet
Implementation-Vendor: Eclipse Foundation
Implementation-Vendor-Id: org.eclipse
Implementation-Version: 5.0.0-SNAPSHOT
Import-Package: jakarta.servlet,jakarta.servlet.annotation,jakarta.servl
 et.descriptor
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Specification-Vendor: Eclipse Foundation
Specification-Version: 5.0
Tool: Bnd-4.0.0.201805111645
```

api/pom.xml
joakime: I just checked the master branch of jsp-api, jaxb-api, and cdi specs.
They all target `<release>1.8</release>` on their compile.

gregw: So is EE9 is java8 byte code.... and this PR excludes module-info.java if built with java8, then this is a noop?  Or am I miss reading this?

joakime: There were 2 `maven-compiler-plugin` entries in this PR when it was first submitted.

1. First one did a normal bytecode compile on release 8 without `module-info.java`.
2. Next one did a compile that only included `module-info.java` using release 9.

The later `package` step bundled them up into a single jar.

api/src/main/java/module-info.java
Thihup: Probably. I'm not sure how to proceed right now. 
Could you please give me a hint in what I should look next?

api/pom.xml
Thihup: I haven't changed the build configuration since the first time I've submitted.
It's actually pretty close to what is done in [JPA api build](https://github.com/eclipse-ee4j/jpa-api/blob/master/api/pom.xml#L254)

api/src/main/java/module-info.java
gregw: yeah, it's not something that you can do on your own here.  It is a policy decision that has to be made at least by the project, but more likely by the EE community as a whole.   So I think it's a bit of a road block, but unfortunately due to the nature of JPMS modules, we really don't want to get it wrong and then try to modify later.  The first published non auto module name needs to be pretty correct in content and intent.    

@stuartwdouglas @markt-asf @bshannon @dblevins any thoughts on this one?

bshannon: You shouldn't export resources that aren't defined by the spec.
The spec doesn't define a resource location for the schemas.

Issue #309, opened 2020-04-06
Clarify methods under various dispatch types.


gregw:
While considering #308 and some jetty TCK failures with getHttpServletMapping, it became apparent that there are some inconsistent and undefined behaviours under various dispatch types.
In the summary below I have indicated if under a given dispatch type if a method should return values for:
 - **target** The target of the current dispatch, ie the Servlet that a dispatch is to.
 - **source** The source of the current dispatch, ie the Servlet that was the target when the dispatch was done or when the async cycle was started.
 - **target^source** Merged values of both the source and target, with target having precedence.
 - **target||source** Target value if there is one, else the source value.

  | REQUEST | FORWARD | INCLUDE | ASYNC | ERROR
-- | -- | -- | -- | -- | --
getRequetURI | target | target | source | target | target
get(Context\|Servlet)Path(Info)? | target | target | source | target | target
getParameter | target | target^source | target^source | target^source | target
getQueryString | target | target\|\|source[1] | source | target\|\|source | target
getHttpServletMapping | target | target | source | source[2] | target
isUserInRole(String) | target | target | target[4] | target | target
getAttribute("javax.servlet.forward.*") |   | source |   |   |  
getAttribute("javax.servlet.include.*") |   |   | target |   |  
getAttribute("javax.servlet.async.*) |   |   |   | source |  
getAttribute("javax.servlet.async.mapping") |   |   |   | source[3] |  
sendError(); getAttribute("javax.servlet.error.*") | target | target | target[5] | target | target
getSession | target | target | target | target | target
getPathTranslated | target | target | target[6] | target | target

Mostly, (other than the bizarre behaviour of include that I never understood and probably lost in ancient history) this table looks pretty sensible.   However there are a number of cases that I think need to be clarified:

**[1]** As per the discussion in #308, the spec doesn't really say what the query string should be during a `FORWARD`, only what the parameters should be.  Jetty currently merges the query string of source and target, while tomcat and undertow use the target if there is one else the source.   
 
[...]

PR #313, opened 2020-04-30
Async Dispatch HttpServletMapping

gregw:
For issue [2] of  #309

Reword the javadoc to align `getHttpServletMapping` under async dispatch to be similar to a `Dispatcher.forward(...)`.  The wording in this section is a bit cumbersome, but I've tried to add clarity without changing it too much.

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
@@ -166,26 +166,47 @@ public interface HttpServletRequest extends ServletRequest {
     public int getIntHeader(String name);
 
     /**
+     * Return the HttpServletMapping for the current dispatch of the request.
      * <p>
-     * Return the {@link HttpServletMapping} by which the {@link HttpServlet} for this {@code HttpServletRequest} was
-     * invoked. The mappings for any applicable {@link jakarta.servlet.Filter}s are not indicated in the result.
-     * If the currently active {@link jakarta.servlet.Servlet} invocation was obtained by a call to
-     * {@link ServletRequest#getRequestDispatcher} followed by a call to {@link RequestDispatcher#forward}, the returned
-     * {@code HttpServletMapping} is the one corresponding to the path used to obtain the {@link RequestDispatcher}.
-     * If the currently active {@code Servlet} invocation was obtained by a call to
-     * {@link ServletRequest#getRequestDispatcher} followed by a call to {@link RequestDispatcher#include}, the returned
-     * {@code HttpServletMapping} is the one corresponding to the first {@code Servlet} in the invocation sequence that
-     * was invoked.
-     * If the currently active {@code Servlet} invocation was obtained by a call to
-     * {@link jakarta.servlet.AsyncContext#dispatch} or {@link jakarta.servlet.AsyncContext#dispatch(String)},
-     * the returned {@code HttpServletMapping} is the one corresponding to the first {@code Servlet} that will be invoked.
-     * See {@link jakarta.servlet.RequestDispatcher#FORWARD_MAPPING},
-     * {@link jakarta.servlet.RequestDispatcher#INCLUDE_MAPPING} and {@link jakarta.servlet.AsyncContext#ASYNC_MAPPING} for
-     * additional request attributes related to {@code HttpServletMapping}. If the currently active {@code Servlet}
-     * invocation was obtained by a call to {@link jakarta.servlet.ServletContext#getNamedDispatcher}, the returned
-     * {@code HttpServletMapping} is the one corresponding to the path for the mapping last applied to this request.
+     * The mapping returned depends on the current {@link jakarta.servlet.DispatcherType} as obtained from
+     * {@link #getDispatcherType()}:
+     *  <dl>
+     *   <dt>{@link jakarta.servlet.DispatcherType#REQUEST}</dt>
+     *   <dd>Return the mapping for the current {@code Servlet}.</dd>
+     *
+     *   <dt>{@link jakarta.servlet.DispatcherType#FORWARD}</dt>
+     *   <dd>If the {@link jakarta.servlet.RequestDispatcher} called was obtained
+     *       by {@link jakarta.servlet.ServletContext#getNamedDispatcher(String)} then
+     *       return the value that would have been returned prior to the call to 
+     *       {@link RequestDispatcher#forward(ServletRequest, ServletResponse)} 
+     *       i.e the value returned is unchanged by the forward call to a named dispatcher.
+     *       Otherwise return the mapping for the target of the dispatch 
+     *       i.e. the mapping for the current {@code Servlet}.</dd>
+     *
+     *   <dt>{@link jakarta.servlet.DispatcherType#INCLUDE}</dt>
+     *   <dd>Return the value that would have been returned prior to the call to 
+     *       {@link RequestDispatcher#include(ServletRequest, ServletResponse)}
+     *       i.e the value returned is unchanged by the include call.</dd>
+     *
+     *   <dt>{@link jakarta.servlet.DispatcherType#ASYNC}</dt>
+     *   <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
+     *       {@code Servlet}.</dd>
+     *
+     *   <dt>{@link jakarta.servlet.DispatcherType#ERROR}</dt>
+     *   <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
+     *       {@code Servlet}.</dd>
+     *  </dl>
      * </p>
-     * 
+     * <p>
+     * Note: For dispatch types that return the mapping for the source of the dispatch, the lookup of the
+     * mapping for the dispatch source is recursive. For example:
+     *  <ul>
+     *   <li>For a sequence Servlet1-&gt;include-&gt;Servlet2-&gt;include-&gt;Servlet3, a call to this method in
+     *       Servlet3 will return the mapping for Servlet1.</li>
+     *   <li>For a sequence Servlet1-&gt;async-&gt;Servlet2-&gt;include-&gt;Servlet3, a call to this method in
+     *       Servlet3 will return the mapping for Servlet2.</li>
+     *  </ul>
+     * </p>
      * <p>
      * The returned object is immutable. Servlet 4.0 compliant implementations must override this method.
      * </p>
--- a/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
@@ -168,18 +168,18 @@ public interface HttpServletRequest extends ServletRequest {
     /**
      * <p>
      * Return the {@link HttpServletMapping} by which the {@link HttpServlet} for this {@code HttpServletRequest} was
-     * invoked. The mappings for any applicable {@link jakarta.servlet.Filter}s are not indicated in the result. If the
-     * currently active {@link jakarta.servlet.Servlet} invocation was obtained by a call to
+     * invoked. The mappings for any applicable {@link jakarta.servlet.Filter}s are not indicated in the result.
+     * If the currently active {@link jakarta.servlet.Servlet} invocation was obtained by a call to
      * {@link ServletRequest#getRequestDispatcher} followed by a call to {@link RequestDispatcher#forward}, the returned
-     * {@code
-     * HttpServletMapping} is the one corresponding to the path used to obtain the {@link RequestDispatcher}. If the
-     * currently active {@code Servlet} invocation was obtained by a call to {@link ServletRequest#getRequestDispatcher}
-     * followed by a call to {@link RequestDispatcher#include}, the returned {@code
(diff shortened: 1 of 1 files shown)


markt-asf:
I agree that the original text is very cumbersome. I've looked at this issue multiple times and each time I have spent a considerable amount of time trying to unpick the intended meaning from the wording used in the original and I still think it is ambiguous.
While I'm generally in favour of keeping changes to the minimum necessary, I think this is a case where it could be worth a clearer re-write. How about the following:
```
* Return the {@link HttpServletMapping} by which the {@link HttpServlet} for this
* {@code HttpServletRequest} was invoked. The mappings for any applicable
* {@link jakarta.servlet.Filter}s are not indicated in the result.
* <p>
* If the currently active {@code Servlet} invocation was obtained by a call to
* {@link ServletRequest#getRequestDispatcher} followed by a call to
* {@link RequestDispatcher#include}, the returned {@code HttpServletMapping} must be
* the same as if this method was called from the including {@code Servlet}.
* <p>
* If the currently active {@link Servlet} invocation was obtained by a call to
* {@link ServletRequest#getRequestDispatcher} followed by a call to
* {@link RequestDispatcher#forward}, the returned {@code HttpServletMapping} must
* reflect the path used to obtain the {@link RequestDispatcher} unless the
* {@link RequestDispatcher} was obtained via
* {@link ServletContext.getNamedDispatcher(String)}, in which case the returned
* {@code HttpServletMapping} must be the same as if this method was called from the
* including {@code Servlet}.
* <p>
* If the currently active {@code Servlet} invocation was obtained by a call to
* {@link jakarta.servlet.AsyncContext#dispatch} or
* {@link jakarta.servlet.AsyncContext#dispatch(String)}, the returned
* {@code HttpServletMapping} reflect the path used to perform the asynchronous
* dispatch.
* <p>
* See {@link jakarta.servlet.RequestDispatcher#FORWARD_MAPPING},
* {@link jakarta.servlet.RequestDispatcher#INCLUDE_MAPPING} and
* {@link jakarta.servlet.AsyncContext#ASYNC_MAPPING} for
* additional request attributes related to {@code HttpServletMapping}.
```

gregw:
@markt-asf sorry but that doesn't read well either.   If we are going to reword it entirely, why not just use the `DispatcherType` in the description (like in the table in the issue) rather than a verbose description of how we got here.   I'll give that a go... stand by....

gregw:
How about:
```java
     * The mapping returned depends on the current {@link jakarta.servlet.DispatcherType} as obtained from
     * {@link #getDispatcherType()}:
     * <dl>
     *     <dt>{@link jakarta.servlet.DispatcherType#REQUEST}</dt>
     *     <dd>Return the mapping for the first {@code Servlet} in the current invocation.</dd>
     *
     *     <dt>{@link jakarta.servlet.DispatcherType#FORWARD}</dt>
     *     <dd>Return the mapping for the first {@code Servlet} in the current invocation.
     *     The mapping for the first {@code Servlet} in the invocation that called
     *     {@link RequestDispatcher#forward(ServletRequest, ServletResponse)} can be obtained
     *     from {@link #getAttribute(String)} using {@link RequestDispatcher#FORWARD_MAPPING} as the key.</dd>
     *
     *     <dt>{@link jakarta.servlet.DispatcherType#INCLUDE}</dt>
     *     <dd>Return the mapping for the first {@code Servlet} in the invocation that called
     *     {@link RequestDispatcher#include(ServletRequest, ServletResponse)}.
     *     The mapping for the first {@code Servlet} in the current invocation can be obtained from
     *     from {@link #getAttribute(String)} using {@link RequestDispatcher#INCLUDE_MAPPING} as the key.</dd>
     *
     *     <dt>{@link jakarta.servlet.DispatcherType#ASYNC}</dt>
     *     <dd>Return the mapping for the first {@code Servlet} in the current invocation.
     *     The mapping for the first {@code Servlet} in the invocation that called
     *     {@link ServletRequest#startAsync()} can be obtained
     *     from {@link #getAttribute(String)} using {@link jakarta.servlet.AsyncContext#ASYNC_MAPPING} as the key</dd>
     *

     *     <dt>{@link jakarta.servlet.DispatcherType#ERROR}</dt>
     *     <dd>Return the mapping for the first {@code Servlet} in the current invocation.
     *     The mapping for the first {@code Servlet} in the invocation that called
     *     {@link HttpServletResponse#sendError(int)} or called {@link HttpServletResponse#sendError(int, String)} or
     *     that threw an exception can be obtained
     *     from {@link #getAttribute(String)} using {@link RequestDispatcher#ERROR_MAPPING} as the key.</dd>
     * </dl>
```

Note  that  currently there is no `RequestDispatcher#ERROR_MAPPING` defined.

markt-asf:
I like the simplification of using the `DispatcherType`.

I'm less sure about the phrase `the first {@code Servlet} in the current invocation`. I'm thinking about when there is a sequence of includes and forwards. I also think the description for error should be the target servlet, not the dispatching servlet.

My turn to try drafting something...

markt-asf:
Ignore that comment about the error case. That is me getting mixed up between the return value and the attribute (which I think might be a case for not discussing the attributes here). I'm still trying to come up with some better language.

gregw:
So we should ignore the last {@code comment} in the discussion that casued the current interaction!   Got it :)

markt-asf:
Right then. How about this:
```
* The mapping returned depends on the current {@link jakarta.servlet.DispatcherType} as obtained from
* {@link #getDispatcherType()}:
* <dl>
*     <dt>{@link jakarta.servlet.DispatcherType#REQUEST}</dt>
*     <dd>Return the mapping for the current {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#FORWARD} (not via a named dispatcher)</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#FORWARD} (via a named dispatcher)</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} where the RequestDispatcher was obtained.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#INCLUDE}</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} where the RequestDispatcher was obtained.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#ASYNC}</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#ERROR}</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
* </dl>
*
* Note that the mapping for the source of a dispatch must also be obtained using the rules
* described above. This means, for example, if Servlet1 includes Servlet2 which in turn
* includes Servlet3, a call to this method in Servlet3 will return the mapping for Servlet1.
```
Once we are happy with the wording for this I suggest we copy/paste to to `getContextPath()` and friends with appropriate replacement of references to `mapping` for each.

stuartwdouglas:
I like the most recent version, I think it is much clearer. 

The only bit I am not sure about is the bottom bit:

> Note that the mapping for the source of a dispatch must also be obtained using the rules
> described above. This means, for example, if Servlet1 includes Servlet2 which in turn
> includes Servlet3, a call to this method in Servlet3 will return the mapping for Servlet1.

Maybe it makes more sense to phrase it in terms of updating the mapping type:

> Note that dispatch types that return the source of the request (namely include and named forwards) must not change the return value of this method. If include is called multiple times
> the returned value must be the same as it was when the first include was performed, until a dispatch is done that updates the mapping. This means, for example, if Servlet1 includes Servlet2 which in turn  includes Servlet3, a call to this method in Servlet3 will return the mapping for Servlet1.

gregw:
@markt-asf  That is is much better, but a problems.

Is it really "where the RequestDispatcher was obtained"?    I have seen applications that obtain a `RequestDispatcher` in init and then reuse it over and over.  So I think it should be:
``` java
/**
* ...
*     <dt>{@link jakarta.servlet.DispatcherType#FORWARD} (via a named dispatcher)</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} when 
*     {@link jakarta.servlet.DispatcherType#forward(ServletRequest, ServletResponse} was called.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#INCLUDE}</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} when 
*     {@link jakarta.servlet.DispatcherType#forward(ServletRequest, ServletResponse} was called.</dd>

```

markt-asf:
Thanks for the feedback. Updated draft follows.
```
* The mapping returned depends on the current {@link jakarta.servlet.DispatcherType} as obtained from
* {@link #getDispatcherType()}:
* <dl>
*     <dt>{@link jakarta.servlet.DispatcherType#REQUEST}</dt>
*     <dd>Return the mapping for the current {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#FORWARD} (not via a named dispatcher)</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#FORWARD} (via a named dispatcher)</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} where
*     {@link RequestDispatcher#forward(ServletRequest, ServletResponse} was called.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#INCLUDE}</dt>
*     <dd>Return the mapping for the source of the dispatch i.e. the mapping for the
*     {@code Servlet} where
*     {@link RequestDispatcher#forward(ServletRequest, ServletResponse} was called.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#ASYNC}</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
*
*     <dt>{@link jakarta.servlet.DispatcherType#ERROR}</dt>
*     <dd>Return the mapping for the target of the dispatch i.e. the mapping for the current
*     {@code Servlet}.</dd>
* </dl>
*
* Note: For dispatch types that return the mapping for the source of the dispatch, the lookup of the
* mapping for the dispatch source is recurrsive. For example:
* <ul>
* <li>For a sequence Servlet1->include->Servlet2->include->Servlet3, a call to this method in
* Servlet3 will return the mapping for Servlet1.</li>
* <li>For a sequence Servlet1->async->Servlet2->include->Servlet3, a call to this method in
* Servlet3 will return the mapping for Servlet2.</li>
```

stuartwdouglas:
LGTM, one comment though, what do you think about changing:

> i.e. the mapping for the {@code Servlet} where {@link RequestDispatcher#forward(ServletRequest, ServletResponse} was called.

To

> i.e. the value that this method would return at the point where {@link RequestDispatcher#forward(ServletRequest, ServletResponse} was called, so the value that is returned by this method remains unchanged by the forward call.

I know it is covered by the bit down the bottom, but maybe this is clearer?

gregw:
I'm OK with @stuartwdouglas more explicit text, however `s/where/when/` because the values return change over time, not with calling location.  
I'll update this PR now and we can review against an actual diff....

gregw:
I committed what had been discussed in ffd266d, but then with some feedback from @janbartel we came up with what we this is a lot simpler.  That is now pushed to this PR and formats as follows:
The mapping returned depends on the current DispatcherType as obtained from ServletRequest.getDispatcherType():

```
DispatcherType.REQUEST, DispatcherType.ASYNC, DispatcherType.ERROR
    Return the mapping for the target of the dispatch i.e. the mapping for the current Servlet.

DispatcherType.INCLUDE
    Return the mapping as prior to the current dispatch. i.e the mapping returned is unchanged 
    by a call to RequestDispatcher.include(ServletRequest, ServletResponse).

DispatcherType.FORWARD
  Return the mapping for the target of the dispatch i.e. the mapping for the current Servlet, unless 
  the RequestDispatcher was obtained via ServletContext.getNamedDispatcher(String), in which 
  case return the mapping as prior to the current dispatch. i.e the mapping returned is changed 
  during a call to RequestDispatcher.forward(ServletRequest, ServletResponse) only if the dispatcher 
  is not a named dispatcher.

For example:
  + For a sequence Servlet1 --include--> Servlet2 --include--> Servlet3, a call to this method in Servlet3 will return the mapping for Servlet1. 
  + For a sequence Servlet1 --async--> Servlet2 --named-forward--> Servlet3, a call to this method in Servlet3 will return the mapping for Servlet2.

The returned object is immutable. Servlet 4.0 compliant implementations must override this method.
```

stuartwdouglas:
This LGTM, if other people are happy with it we should merge.

Do we have a policy on squashing commits? IMHO this should be squashed down to a single commit, to make the repo history cleaner.

gregw:
+1 on squashing

markt-asf:
LGTM.

stuartwdouglas:
@gregw do you want to manually squash, or do you want me to do it from the UI?

gregw:
@stuartwdouglas I did the squash and merge via the UI.

PR #319, opened 2020-05-30
javax --> jakarta specification

thadumi:
Signed-off-by: thadumi <th.theodor.th@gmail.com>

I prepared a draft of the specification with the changes from javax to jakarta. 

I removed the section 1.6 according to the discussion https://www.eclipse.org/lists/servlet-dev/msg00180.html


spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -277,8 +271,7 @@ the servlet under the compliance with RFC 7234.
 
 A servlet container may place security
 restrictions on the environment in which a servlet executes. In a Java
-Platform, Standard Edition (J2SE, v.1.3 or above) or Java Platform,
-Enterprise Edition (Java EE, v.1.3 or above) environment, these
+Platform, Standard Edition (J2SE, v.1.3 or above), these
```
stuartwdouglas, line 274: Should we change this to be 1.8 or above?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1387,9 +1370,9 @@ methods of the `ServletRequest` interface:
 Only one attribute value may be associated
 with an attribute name.
 
-Attribute names beginning with the prefixes of
-`java.` and `javax.` are reserved for definition by this specification.
-Similarly, attribute names beginning with the prefixes of `sun.` ,
+Attribute names beginning with the prefix of
+`jakarta.` are reserved for definition by this specification.
+Similarly, attribute names beginning with the prefixes of `java.`, `sun.` ,
 `com.sun., oracle and com.oracle` are reserved for definition by Oracle
```
stuartwdouglas, line 1376: Do we still need this bit? I don't know if we should still reserve the Sun and Oracle namespaces.

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: I'd remove references to Java versions here and just say the security restrictions used by the Java platform may be used.

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: I'd say no.

gregw: I don't think we should say anything about `java.`, `sun.`, `com.sun,`, `oracle` and `com.oracle`.

However, I think it may be a frequent porting mistake to forget to change the `javax.` names (for those not using the constants).   We could do an IAE to catch gets of the old names?  Or maybe a note to say that `Containers may throw IllegalArgumentException for unsupported attributes (eg. old javax. attributes` ?

stuartwdouglas: I would prefer to say nothing, so implementation could add their own backwards compatibility handling.

thadumi: Thanks for the discussion. Indeed since we are going to [empty also the change-log](https://www.eclipse.org/lists/servlet-dev/msg00183.html), it will be confusing to mention `javax. *` which refer to the past mostly.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -338,39 +331,25 @@ independence.
...
+of the JakartaEE Platform, 9 footnote:[Please see the JakartaEE
+Platform, specification available at
+`https://jakarta.ee/specifications/platform/9/`].
 Servlet containers and servlets deployed into them must meet additional
-requirements, described in the Java EE specification, for executing in a
-Java EE environment.
+requirements, described in the Jakarta EE specification, for executing in a
+Jakarta EE environment.
 
-=== [[a384]] Compatibility with Java Servlet Specification Version 2.5
```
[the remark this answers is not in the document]
thadumi: Yes, I agree. As discussed in https://www.eclipse.org/lists/servlet-dev/msg00183.html cleaning the change-log would be the best solution.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -68,30 +62,30 @@ such.
 === API Reference
 
 The full specifications of classes, interfaces,
-and method signatures that define the Java Servlet API, as well as their
+and method signatures that define the Jakarta Servlet API, as well as their
 accompanying Javadoc™ documentation, is available online at
-`http://docs.oracle.com/javaee/` .
+`https://jakarta.ee/specifications/servle` .
```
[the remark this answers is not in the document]
thadumi: Thanks

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -9097,1024 +9097,14 @@ for Dependency Injection" of the JakartaEE Platform 9 specification.
...
-Clarification of SERVER_NAME and SERVER_PORT
-in getServerName() and getServerPort() (14.2.16)
-
-Clarification: “run-as” identity must apply
-to all calls from a servlet including init() and destroy() (12.7)
-
-Login/logout description and methods added
-(12.10, 15.1.7)
-
+=== [[a384]] Compatibility with Jakarta Servlet Specification Version 4.0
```
thadumi, line 9100: @gregw , please let me know what you thing about this solution.
 
Also I tried to find the rules for the anchor numbers but I didn't find them. Thus, if I need to change the numbers could you explain me how they are chosen?

gregw: I think we can be stronger in this description.  5.0 is ONLY a change in the namespace, with no other changes.... at least few other changes... so few that if we do some they should be explicitly listed here.

arjantijms:
Maybe at some point we want to split it up into chapters, but that point in time is not here.

PR #321, opened 2020-06-11
HTTP/2 fix

senivam:

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServlet.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServlet.java
@@ -75,8 +75,6 @@ public abstract class HttpServlet extends GenericServlet {
     private static final String METHOD_PUT = "PUT";
     private static final String METHOD_TRACE = "TRACE";
 
-    private static final String PROTOCOL_VERSION_1_0 = "1.0";
-
     private static final String HEADER_IFMODSINCE = "If-Modified-Since";
     private static final String HEADER_LASTMOD = "Last-Modified";
 
@@ -145,11 +143,7 @@ public abstract class HttpServlet extends GenericServlet {
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         String protocol = req.getProtocol();
         String msg = lStrings.getString("http.method_get_not_supported");
-        if (!protocol.endsWith(PROTOCOL_VERSION_1_0)) {
-            resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
-        } else {
-            resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
-        }
+        switchProtocol(protocol, resp, msg);
     }
 
     /**
@@ -250,11 +244,7 @@ public abstract class HttpServlet extends GenericServlet {
     protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         String protocol = req.getProtocol();
         String msg = lStrings.getString("http.method_post_not_supported");
-        if (!protocol.endsWith(PROTOCOL_VERSION_1_0)) {
-            resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
-        } else {
-            resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
-        }
+        switchProtocol(protocol, resp, msg);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -143,7 +145,7 @@ public HttpServlet() {
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         String protocol = req.getProtocol();
         String msg = lStrings.getString("http.method_get_not_supported");
-        if (protocol.endsWith("1.1")) {
+        if (!protocol.endsWith(PROTOCOL_VERSION_1_0)) {
```
gregw, line 148: I know that HTTP 0.9 is really meant to be supported anymore, but there are probably still some usages out there.
Perhaps a switch approach is better:
```suggestion
        switch(protocol) {
          case "HTTP/0.9":
          case "HTTP/1.0":
            resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
            break;
          default:
            resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
            break;
        }
```

senivam: yes, good point, I've modified changes accordingly

m0mus:
@gregw This is a blocking PR for work we are doing. Can you please take a look and merge it if it looks good. We also need to release a new version of *javax* based servlet API containing this fix. In order to do it we will need to commit it to *master* as well as in other *javax* branch. Can you please suggest what branch shall we use for that?

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -143,11 +143,7 @@ public HttpServlet() {
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         String protocol = req.getProtocol();
         String msg = lStrings.getString("http.method_get_not_supported");
-        if (protocol.endsWith("1.1")) {
-            resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
-        } else {
-            resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
-        }
+        switchProtocol(protocol, resp, msg);
```
gregw, line 146: I don't like the method name... it is kind of unclear what it does.   I think the following approach would be better:
```suggestion
        resp.sendError(getMethodNotSupportedCode(protocol), msg);
```

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -317,10 +305,18 @@ protected void doPut(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+    private static void switchProtocol(String protocol, HttpServletResponse resp, String msg) throws IOException {
+        switch (protocol) {
+        case "HTTP/0.9":
+        case "HTTP/1.0":
             resp.sendError(HttpServletResponse.SC_BAD_REQUEST, msg);
+            break;
+        default:
+            resp.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, msg);
+            break;
         }
```
gregw, line 320: Make non-static protected so that it can be overridden
return the code rather than apply it.
```suggestion
    protected int getMethodNotSupportedCode(String protocol) {
        switch (protocol) {
          case "HTTP/0.9":
          case "HTTP/1.0":
            return HttpServletResponse.SC_BAD_REQUEST;
        default:
            return HttpServletResponse.SC_METHOD_NOT_ALLOWED;
        }
```

gregw:
@m0mus sorry but I personally would like a few more changes, plus as this is a code change it would be good to get review from one of the leads (@markt-asf @stuartwdouglas ) before it is merged.

m0mus:
> @m0mus sorry but I personally would like a few more changes, plus as this is a code change it would be good to get review from one of the leads (@markt-asf @stuartwdouglas ) before it is merged.

There is nothing to sorry about. We will address your comments and I added both Mark and Stuart as reviewers.

senivam:
OK, so I did requested changes

gregw:
This LGTM, but I'd like another committer to agree before merging.

markt-asf:
LGTM +1 to merge

senivam:
Thank you for accepting the changes. Don't you mind if new version of the API will be released (to jakarta oss staging) which will include those changes?

PR #329, opened 2020-07-03
Spec review part1

markt-asf:
Reviewed references to Java, removed from old JCP/EG references, "JakartaEE" -> "Jakarta EE", version number updates.

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -92,7 +92,7 @@ Platform, Web site: `https://jakarta.ee/specifications/`
 === [[a26]] Other Important References
 
 The following Internet specifications provide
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1371,9 +1336,7 @@ with an attribute name.
 
 Attribute names beginning with the prefix of
 `jakarta.` are reserved for definition by this specification.
```
gregw, line 1338: Isn't "this specification" only going to define `jakarta.servlet` attributes?

markt-asf: As far as I recall all attributes will be `jakarta.servlet.*`. That said, I don't see the harm in reserving the wider namespace.

PR #331, opened 2020-07-08
Fix typo auth-contraint -> auth-constraint

Thihup:
Signed-off-by: Thiago Henrique Hüpner <thihup@gmail.com>


arjantijms:
Thanks, good catch! :)

PR #337, opened 2020-07-18
Spec pass 3

markt-asf:
Proof-reading. Mainly fixing typos and trying to make formatting consistent.


arjantijms:
LGTM, small doubt is whether it should not still be "Servlet Container Profile" instead of "servlet Container Profile".

arjantijms:
The name "JASPIC" also isn't used anymore, so that would just be "Jakarta Authentication". We can change that in a next PR or this one.

markt-asf:
We can do another pass to clean up those issues. Is there such a thing as the Servlet Container Profile? I don't recall such a thing at the platform level. I'm tempted to say we need to reword that but I'll find it the spec so I can look at it in context.
JASPIC - sure, will do. Is there a similar change required for JACC?

arjantijms:
> Is there such a thing as the Servlet Container Profile?

yes, but its in the context of Jakarta Authentication (previously JASPIC), see here: 

https://jakarta.ee/specifications/authentication/2.0/authentication-spec-2.0-SNAPSHOT.html#a365

> JASPIC - sure, will do. Is there a similar change required for JACC?

Yes, JACC became Jakarta Authorization.

I think there's an EJB reference lurking too in the Servlet spec, which should become Jakarta Enterprise Beans.

markt-asf:
Got it. I'll do a 4th pass.

PR #338, opened 2020-07-19
Spec pass 4

markt-asf:
Correct a few things in the previous pass and a couple of additional issues that have been identified.

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -6972,9 +6972,9 @@ developers, system administrators, and system integrators).
 
 To facilitate portable implementation and
 integration of additional container authentication mechanisms, it is
-recommended that all servlet containers implement the servlet Container
-Profile of The Jakarta Authentication SPI for Containers.
-The SPI is available for download at:
+recommended that all servlet containers implement the Servlet Container
+Profile of The Jakarta Authentication Specification.
+The specification is available for download at:
 https://jakarta.ee/specifications/authentication/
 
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -587,13 +587,13 @@ becomes available, the thread handling that event continues processing
...
-Jakarta EE features such as
-<<_Web_Application_Environment>> and
-<<Propagation of Security Identity in EJB™ Calls>> are available only to
-threads executing the initial request or when the request is dispatched
-to the container via the `AsyncContext.dispatch` method. Jakarta EE features
-may be available to other threads operating directly on
-the response object via the `AsyncContext.start(Runnable)` method.
+Jakarta EE features such as <<_Web_Application_Environment>> and
+<<Propagation of Security Identity in Jakarta Enterprise Bean Calls>>
+are available only to threads executing the initial request or when
```
gregw, line 592: Does "initial request" here include `RequestDispatcher.forward` and `RequestDispatcher.include`?   I would think so, but we should probably say explicitly or at least with broader language that doesn't feel like "initial dispatch".   Also any error dispatches are not the initial request, but I also think EE features are available.   

But more importantly for asynchronous processing, there are also async IO callbacks: onDataAvailable, onWritePossible, onAllDataRead and onError that should also have EE features enabled in them.

Rather than enumerating APIs, would it not be better just to say that all container dispatched threads associated with a request should have the EE features enabled?

arjantijms:
>  I think we have more work to do there to improve clarity.

That may have to be done for Servlet 5.1 or whatever new version, as the deadline for Servlet 5 is today.

spec/src/main/asciidoc/servlet-spec-body.adoc
arjantijms: Greg, if you want, can you open a new issue for that?

PR #356, opened 2020-09-08
Issue #225 doHead contentLength

### LINKED ISSUE #225: HttpServlet.doHead sets content-length
The implementation of [HttpServlet.doHead](https://github.com/eclipse-ee4j/servlet-api/blob/master/src/main/java/javax/servlet/http/HttpServlet.java#L192-L197) sets content-length, even if the `doGet` implementation does not.  This results in different headers being sent, specifically a GET request that would send a chunked response will respond to a HEAD request without being chunked.  This is contrary to [RFC7230](https://tools.ietf.org/html/rfc7231#section-4.3.2) which allows for content headers to be omitted, but not altered.

gregw:
Fix #225 by:
 + converting the contentLength field to a long
 + checking against bufferSize before setting a content length to simulate a buffer overflow
 + support flush and close
 + pass through async IO listener

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServlet.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServlet.java
@@ -610,17 +610,16 @@ class NoBodyResponse extends HttpServletResponseWrapper {
 
     // file private
     void setContentLength() throws IOException {
-        if (!didSetContentLength) {
+        // Set the content length IFF it is not already set and the app the output stream
+        // has been created.
+        if (!didSetContentLength && noBody != null) {
             if (writer != null) {
-                writer.flush();
+                noBody.flush(writer);
             }
-            if (noBody != null) {
-                long contentLength = noBody.getContentLength();
-                if (contentLength > getBufferSize() && !isCommitted()) {
-                    flushBuffer(); // commit without content-length
-                } else {
-                    setContentLengthLong(noBody.getContentLength());
-                }
+
+            // If the response has not been committed, the content length can be set.
+            if (!isCommitted()) {
+                setContentLength(noBody.getContentLength());
             }
         }
     }
@@ -708,7 +707,11 @@ class NoBodyOutputStream extends ServletOutputStream {
 
     private final NoBodyResponse response;
     private final ServletOutputStream wrapped;
-    private long contentLength = 0;
+
+    // The real contentLength can a long, but this class tracks it only whilst less than
+    // the buffer size, which is an int
+    private int contentLength = 0;
+    private boolean writerFlush = false;
 
     // file private
     NoBodyOutputStream(NoBodyResponse response) throws IOException {
@@ -717,13 +720,13 @@ class NoBodyOutputStream extends ServletOutputStream {
     }
 
     // file private
-    long getContentLength() {
+    int getContentLength() {
         return contentLength;
     }
 
     @Override
-    public void write(int b) {
-        contentLength++;
+    public void write(int b) throws IOException {
+        incrementContentLength(1);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


stuartwdouglas:
I can't help thinking that it would be better to just have doHead delegate to doGet, and leave this all up to the container. 

This implementation would not actually work in undertow, as the delegated isCommited would only return true if bytes had actually been written to the wire, which they haven't in this case. Conversely delegating directly to doGet would result in Undertow always returning the exact same headers.

gregw:
@stuartwdouglas that's what Jetty does with it's own servlets.    But we've had a minimal implementation since dot, so probably should keep it for any light weight implementations that don't do their own HEAD handling.

So perhaps we need to check against buffer size on every write and as soon as content written exceeds the buffer size, we do a flushBuffers to commit the response?

stuartwdouglas:
That sounds reasonable.

stuartwdouglas (approved):
LGTM, although some comments around writerFlush would not go astray as it took me a little bit to realize what is going on there.

Some tests could also be nice here, but as it would likely involve just mocking everything I don't know useful they would actually be.

gregw:
@stuartwdouglas I reworked the writerFlush mechanism to make it clearer what it is doing.  
I'll look at some tests tomorrow.

However, to really test this we need to expose apps that actually use this style of HEAD response to the change.... how do we do that before releasing?

stuartwdouglas:
Thats why I think all we can really do here is unit style testing work mocks, I don't think we can actually test with a real container (at least not easily). 

Personally I try and avoid using mocks as much as possible, but maybe in this case they actually make sense.

gregw:
@stuartwdouglas I added some unit tests... good thing too as there were problems.  The complexity of this is surprising and makes me worry about this change.  Perhaps being bug for bug compatible is better than trying to fix in the generic class and failing???

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -598,78 +598,135 @@ public void service(ServletRequest req, ServletResponse res) throws ServletExcep
...
+        writer = null;
+        usingOutputStream = false;
+        setContentLength = -1;
+    }
+
+    @Override
+    public void resetBuffer() {
+        super.resetBuffer();
+        flushWriter();
+        setContentLength = -1;
```
stuartwdouglas, line 713: I don't think this is correct, Javadoc says:

`
Clears the content of the underlying buffer in the response without clearing headers or status code. If the response
      has been committed, this method throws an <code>IllegalStateException</code>.
`

So the only way the content length has been set is if the user explicitly set it, and this method should not mess with headers.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -598,78 +598,135 @@ public void service(ServletRequest req, ServletResponse res) throws ServletExcep
...
     // file private
-    void setContentLength() {
-        if (!didSetContentLength) {
-            if (writer != null) {
+    void checkContentLength() throws IOException {
+        // Set the content length IFF it is not already set and the app the output stream
+        // has been created.
+
+        if (!isCommitted()) {
+            if (setContentLength == -1) {
```
stuartwdouglas, line 621: What if the user explicitly set `Transfer-Encoding: chunked` in Undertow's case if the user has explicitly set a TE header we won't mess with the content length.

stuartwdouglas (changes requested):
I think I found another 2 edge cases. I must admit I am feeling less and less confident about this as it gets more complex.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
gregw: Ah yes.
But I think that reveals the folly of this PR.    I don't know how we are going to perfectly mimic the headers that a container might use in GET without actually deferring the handling to the container.
The original idea was to set content-length if a similar GET request would have, but there is no compulsion on a container to set content-length even if it knows it.

Perhaps a better idea would be to provide some way for a container to disable the whole nobody wrapper stuff anyway. That way the container can opt to handle HEAD itself.

gregw:
I'm closing this as I don't have confidence that this will not cause more problems than it resolves.... better the devil we know....

I will instead consider how we can modify to allow an implementation to entirely disable the nobody mechanism and just do it's own HEAD handling.

stuartwdouglas:
+1 to closing it. How big an issue is this in reality? In general there is no requirement for two requests to the same endpoint to return the exact same content, so content-length/encoding may differ between subsequent requests anyway.

PR #359, opened 2020-09-08
Remove empty manifest and readme

pzygielo:
[javaee:README](https://github.com/javaee/servlet-spec/blob/d3d0f3ba4dc9256122f01ebdd4be4fae2a586df3/src/main/resources/META-INF/README#L41-L44) (since 2010)

I took that effort mentioned above already and instead of waiting for the original author to have some time in the future I present this change for consideration today.
I've built api module twice (from current master (20c4e41) and from 539c30b presented in this PR) and here is the diff of manifests from produced jars:
```
6c6
< Bnd-LastModified: 1599586342818
---
> Bnd-LastModified: 1599586039507
```
which I feel is acceptable.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -96,9 +96,6 @@
             </resource>
             <resource>
                 <directory>src/main/resources</directory>
-                <excludes>
-                    <exclude>META-INF/README</exclude>
-                </excludes>
             </resource>
         </resources>
         <pluginManagement>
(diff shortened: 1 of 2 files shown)


arjantijms:
Indeed, wonder if Sahoo will ever find that time, but I'm sure he would be happy to have seen you done this.

arjantijms (approved):

PR #360, opened 2020-09-08
Clarifly the JSR-133 sematics of the init methods

### LINKED ISSUE #81: Clarify to users that the container will ensure work done in Servlet#init(ServletConfig) will be visible to threads which later invoke service methods.
As a longtime user of the Servlet specification, I have recently become more aware of thread-safety concerns, and when making my last servlet, I realised that if doGet() and friends are invoked from a thread pool, they are only guaranteed to see assignments I make in init(ServletConfig) if someone else has taken adequate care that the work done by that thread is visible to others.

Presumably, containers have in fact done this for nearly two decades. But it would be nice to have it spelled out somewhere, so that I can rely on this behaviour, and don't need to second-guess the environment and perform my own locking/synchronization.

stuartwdouglas:
Fixes #81


stuartwdouglas:
@gregw @markt-asf I don't feel strongly about this either way, but I think we should either add some text like this or close the original issue.

martin-g (approved):

gregw (approved):
I'm +0 on this.  The text is good, but a bit over the top.... however, it will close out an issue.
I'm fine with merge and close.... or just close th issue.

arjantijms:
@stuartwdouglas Just to be sure, is this intended for 5.0 or 5.1? There's no milestone set.

stuartwdouglas:
5.1, I have updated the milestone.

PR #361, opened 2020-09-09
Begin work on Servlet 5.1

stuartwdouglas:

### DIFF
--- a/api/src/main/resources/jakarta/servlet/resources/web-app_5_1.xsd
+++ b/api/src/main/resources/jakarta/servlet/resources/web-app_5_1.xsd
@@ -1,0 +1,342 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsd:schema xmlns="http://www.w3.org/2001/XMLSchema"
+            targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified"
+            version="5.1">
+  <xsd:annotation>
+    <xsd:documentation>
+
+      Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
+      
+      This program and the accompanying materials are made available under the
+      terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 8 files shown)


api/src/main/resources/jakarta/servlet/resources/web-app_5_1.xsd
```diff
@@ -0,0 +1,342 @@
...
+            targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified"
+            version="5.1">
+  <xsd:annotation>
+    <xsd:documentation>
+
+      Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
```
stuartwdouglas, line 12: Do we still need this line in the new versions? It is based on the existing work so I am guessing yes.

arjantijms: I'm fairly sure we need to keep that.

Btw, when changing over to 5.1, let's be extra careful TCK jobs etc run against the 5.0 branch, which could still be updated (the release ballot hasn't started yet)

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -278,18 +280,6 @@ independence.
 * They can access the large set of APIs
 available for the Java platform.
 
-=== Relationship to Jakarta EE Platform
```
stuartwdouglas, line 281: I think we should delete this section, as it is likely to be impossible to keep in sync. Servlet 5.0 will be part of the platform spec, but 5.1 is an unknown. It may end up in the next version of the platform, or we maybe have 5.2 by then. A spec may not be part of the platform when it is published, then included later.

arjantijms: +1 The platform should say what and in which versions it includes things.

arjantijms (approved):

markt-asf (approved):
Looks like a reasonable starting point to me.

martin-g (approved):

gregw (approved):
LGTM

arjantijms:
With multiple approvals, and Servlet 5.0 being done, I'll now merge this.

PR #364, opened 2020-09-19
Some more spec document cleanup

### LINKED ISSUE #363: Servlet 5 Spec. Doc. Final Review Comments
I read through the current Spec. that is posted for review at the EE Spec. site. My compliments on the clean-up work you have done, compared to what was originally contributed. It really looks great. Thank you!
These are comments based on the [Ballot PR](https://github.com/jakartaee/specifications/pull/232), The preview spec. rendition ([pdf](https://deploy-preview-232--jakartaee-specifications.netlify.app/specifications/servlet/5.0/jakarta-servlet-spec-5.0.pdf), [html](https://deploy-preview-232--jakartaee-specifications.netlify.app/specifications/servlet/5.0/jakarta-servlet-spec-5.0.html)). The following probably should be considered for revision, but I do not consider these ballot critical issues:

- References in Specs. to optional and pruned specifications should be checked and updated, across all Specs. (see the [EE 9 Release Plan](https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee9/JakartaEE9ReleasePlan) for list of optional and pruned specs). Pruned spec references should be removed. Optional references should indicate that these components are optional.
- The preface still refers to "Reference Implementation." There will be a compatible implementation as part of the initial review, but it should be referred to as a "Compatible Implementation" (or CI).
- Use Jakarta EE Platform TCK, or just TCK, instead of CTS.
- The final draft should not include any 'status' designation. The top level POM notes indicate that <properties>/<status> should be blank when generating the final version (this is only visible in the PDF).
- Legacy acronyms should be revised to use the new Jakarta / EE4J names. JAX-RS should be Jakarta Restful Web Services, JSF should be Jakarta Server Faces, JAX-WS should be Jakarta Enterprise Web Services etc. (In call cases, you can leave out Jakarta).
- References to external specifications probably need to specify a minimum version due to the name-space changes (e.g. Authorization 2.0 or higher)
- In Chapt. 14, Deployment Descriptors the text still says "All legacy deployment descriptors must be supported." I am not sure if that will be possible. Please review and update the text if needed.
- In Chapt 15, subsection 4, Deployment -- There are optional and pruned component references in this section. They should be evaluated and the text should be updated to remove pruned specifications as well as confirming that optional components are indeed optional.
- Chapt 15.5.15 \@WebServiceRefs Annotation -- The reference to JAX-WS Spec. section 7 probably would be better if it included the section name -- just in case it changes.

Lastly, just a reminder, don't forget to update the namespace descriptor documents (https://jakarta.ee/xml/ns/jakartaee/) -- currently they are all in a preliminary status. Once the Spec. has been approved, they should be changed to Final.

stuartwdouglas:
Fixes #363

This should address most of comments in the issue, bar a few things:

> In Chapt. 14, Deployment Descriptors the text still says "All legacy deployment descriptors must be supported." I am not sure if that will be possible. Please review and update the text if needed.

Not sure what to do about this one. The XML does not directly reference package names so really there is no reason why we can't support old versions, but old apps will be using the old namespace which clearly won't work without some form of transformation. Maybe we should just drop this requirement from the spec altogether, and leave it up to the implementations as to what they support (however add it back in 5.1, so 5.1 needs to support 5.0 descriptors).

> References to external specifications probably need to specify a minimum version due to the name-space changes (e.g. Authorization 2.0 or higher)

I was not really sure what to do about this, maybe it needs a new section on compatible spec versions?

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -8,9 +8,9 @@ Copyright (c) 2020 Contributors to the Eclipse Foundation.
 Eclipse is a registered trademark of the Eclipse Foundation. Jakarta
 is a trademark of the Eclipse Foundation. Oracle and Java are
 registered trademarks of Oracle and/or its affiliates. Other names
-may be trademarks of their respective owners. 
+may be trademarks of their respective owners.
 
-The Jakarta Servlet Team - {revdate}
+The Jakarta Servlet Team
 
 Comments to: servlet-dev@eclipse.org
 
@@ -26,14 +26,14 @@ The specification is intended to be a complete
 and clear explanation of the Jakarta Servlet API, but if questions remain, the
 following sources may be consulted:
 
-* A reference implementation (RI) has been made
+* A compatible implementation (CI) has been made
 available which provides a behavioral benchmark for this specification.
 Where the specification leaves implementation of a particular feature
 open to interpretation, implementors may use the reference
 implementation as a model of how to carry out the intention of the
 specification.
 
-* A compatibility test suite (CTS) has been
+* The Jakarta EE Platform TCK has been
 provided for assessing whether implementations meet the compatibility
 requirements of the Jakarta Servlet API standard. The test results have
 normative value for resolving questions about whether an implementation
@@ -88,6 +88,9 @@ referenced throughout this specification:
 
 These specifications may be found at the Jakarta EE
 Platform, web site: `https://jakarta.ee/specifications/`.
+
+References to other specifications that are part of the Jakarta EE platform exclude the versions of
+those specifications that target Jakarta EE 8 due to the different name-space used.
 
 === Other Important References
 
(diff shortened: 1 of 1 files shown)


markt-asf:
I'm leaning towards keeping the requirement to support legacy descriptors. (Automated) migration of applications is simpler if the descriptors can be left as-is and the more we can do to ease the migration path, the more likely it is the users will migrate sooner and the associated costs of retaining this requirement appear to be negligible.

Rather than add versions to every spec reference, what about a general statement along the lines of

> References to other specifications that are part of the Jakarta EE platform exclude the versions of those specifications that target Jakarta EE 8 due to the different name-space used.

stuartwdouglas:
I have updated the section on the spec versions. +1 to leaving the compatibility requirements as is.

hussainnm (changes requested):
JAX-WS is Jakarta XML Web Services and not Jakarta Enterprise Web Services

stuartwdouglas:
I have updated this and rebased my 5.1 PR. IMHO we should get this in, then merge the 5.1 PR and branch 5.0.

arjantijms:
> I have updated this and rebased my 5.1 PR. IMHO we should get this in, then merge the 5.1 PR and branch 5.0.

Indeed, now would be the time for that. I'll branch 5.0, and do the 5.1 PR merge then.

arjantijms (approved):

PR #370, opened 2020-10-12
Fixes #368 Updates javadoc for getRemoteAddr

### LINKED ISSUE #368: Impl Check: ServletRequest.getRemoteAddr and intermediaries
Hello Experts,

I have a customer with a question about [ServletRequest.getRemoteAddr](https://jakarta.ee/specifications/servlet/4.0/apidocs/javax/servlet/ServletRequest.html#getRemoteAddr--).  Consider the following network layout.

```
Browser
   |
  \|/
CloudFlare
   |
  \|/
----Azure-boundary-----
NGINX A
   |
  \|/
NGINX B
   |
  \|/
Tomcat (Spring Cloud)
```

Please confirm that `getRemoteAddr` in the tomcat should return the **CloudFlare** not the **NGINX B**.

Thanks,

Ed

gregw:
I ended up needing to fix a bit more for #368:
 + reference to RFC 7239  for obtaining local/remote IP and ports
 + removed historic references to CGI variables
 + relaxed the dependence on the host header, which may not be present in HTTP/2
 + removed the reference to dotted IP address form (no longer valid for IPv6)

I'm wondering if we should also say anything about [] around IPv6?

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletRequest.java
+++ b/api/src/main/java/jakarta/servlet/ServletRequest.java
@@ -208,9 +208,9 @@ public interface ServletRequest {
 
     /**
      * Returns the host name of the server to which the request was sent. It may be derived from a protocol specific
-     * mechanism, such as the <code>Host</code> header, or the HTTP/2 authority,
-     * or <a href="https://tools.ietf.org/html/rfc7239">RFC 7239</a>, otherwise the
-     * resolved server name or the server IP address.
+     * mechanism, such as the <code>Host</code> header, or the HTTP/2 authority, or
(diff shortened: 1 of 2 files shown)


stuartwdouglas (approved):

markt-asf (approved):
While it does seem odd to have references to CGI variables in the Javadoc, I have had to rely on CGI spec references in the past to determine correct behaviour. I'm happy for the references to be removed on the understanding that there is no intention to change required behaviour.

gregw:
@markt-asf I removed the CGI references as they were written over a decade ago and who knows if CGI has been update/changed/refined in that time... nor given the RFC7239 nuances we have discussed if they will always be the same value.   Thus it felt like it was not worth anything as part of the specification, nor could we really say it was an invariant.

PR #371, opened 2020-10-30
Remove deprecated that return null or empty

Cousjava:
Removes methods that have long been deprecated and are already are specified to either return `null` or an empty `Enumeration` in all cases, so therefore their removal should not affect any applications.


stuartwdouglas:
I am in two minds about this.

On the one hand these methods are clearly just baggage, and we are better off without them. On the other hand the Servlet API is so widely used that any backwards incompatible changes can have a massive cost. I know the rename has broken backwards compatibility anyway, but bytecode manipulation utilities can deal with this to some extent.

I think the main issue that I am worried about is that if you have a ServletContext or HttpSession wrapper with @Override this will no longer compile after these changes. Servlet 5.1 is supposed to be a minor release that focuses on samesite cookies, so maybe we should wait for a 6.0 release? 

We could also start to plan for the removal by adding default method bodies that just return null to these interfaces. This means that any applications that actually implement these interfaces can just ignore these methods completely.

stuartwdouglas:
I am in two minds about this.

On the one hand these methods are clearly just baggage, and we are better off without them. On the other hand the Servlet API is so widely used that any backwards incompatible changes can have a massive cost. I know the rename has broken backwards compatibility anyway, but bytecode manipulation utilities can deal with this to some extent.

I think the main issue that I am worried about is that if you have a ServletContext or HttpSession wrapper with @Override this will no longer compile after these changes. Servlet 5.1 is supposed to be a minor release that focuses on samesite cookies, so maybe we should wait for a 6.0 release?

We could also start to plan for the removal by adding default method bodies that just return null to these interfaces. This means that any applications that actually implement these interfaces can just ignore these methods completely.

gregw:
@stuartwdouglas so the double commenting was one from each of your two minds?   But if so, they are in agreement as they said the same thing.  I agree too :+1:

Cousjava:
Based off the comments here I am replacing it with #372

PR #372, opened 2020-11-02
Provide defaults for deprecated methods

Cousjava:
These methods all already specify that they return either null or
an empty enumeration so they are now provided by default for easier
removal later.

Based off comments in #371

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -2,6 +2,7 @@
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
  * All rights reserved.
  * Copyright 2004 The Apache Software Foundation
+ * Copyright (c) 2020 Payara Services Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +27,7 @@ import java.util.Enumeration;
 import java.util.EventListener;
 import java.util.Map;
 import java.util.Set;
+import java.util.Vector;
 
 /**
  * Defines a set of methods that a servlet uses to communicate with its servlet container, for example, to get the MIME
@@ -361,11 +363,11 @@ public interface ServletContext {
      * shared business logic by invoking methods on common non-servlet classes.
      *
      * @param name the servlet name
-     * @return the {@code jakarta.servlet.Servlet Servlet} with the given name
+     * @return {@code null}
      * @throws ServletException if an exception has occurred that interfaces with servlet's normal operation
      */
     @Deprecated
-    public Servlet getServlet(String name) throws ServletException;
+    default public Servlet getServlet(String name) throws ServletException {return null;}
 
     /**
      * @deprecated As of Java Servlet API 2.0, with no replacement.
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -378,7 +380,7 @@
      * @return an <code>Enumeration</code> of {@code jakarta.servlet.Servlet Servlet}
      */
     @Deprecated
-    public Enumeration<Servlet> getServlets();
+    default public Enumeration<Servlet> getServlets() { return new Vector<Servlet>().elements();}
```
martin-g, line 383: java.util.Collections#emptyEnumeration() ?

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -391,7 +393,7 @@
      * @return an <code>Enumeration</code> of {@code jakarta.servlet.Servlet Servlet} names
      */
     @Deprecated
-    public Enumeration<String> getServletNames();
+    default public Enumeration<String> getServletNames() { return new Vector<String>().elements();}
```
martin-g, line 396: java.util.Collections#emptyEnumeration() ?

api/src/main/java/jakarta/servlet/http/HttpSessionContext.java
```diff
@@ -55,5 +56,5 @@
      *
      */
     @Deprecated
-    public Enumeration<String> getIds();
+    default public Enumeration<String> getIds() {return null; }
```
martin-g, line 59: java.util.Collections#emptyEnumeration() ?

The deprecation message says `This method must return an empty <code>Enumeration</code>`

Cousjava: It also says `@return null`

api/src/main/java/jakarta/servlet/http/HttpSessionContext.java
```diff
@@ -55,5 +58,7 @@
      *
      */
     @Deprecated
-    public Enumeration<String> getIds();
+    default public Enumeration<String> getIds() {
```
stuartwdouglas, line 61: Well this is awesome, we have contradictory Javadoc on the method (the first part says empty enumeration, the @return says null).

I think an empty enumeration would be safer here.

Cousjava: Done and updated the Javadoc as well.

stuartwdouglas (approved):

stuartwdouglas (changes requested):
This LGTM, but I just realised that we should also include a note in the changelog about this.

Cousjava:
@stuartwdouglas Where is the changelog to add the note to?

stuartwdouglas:
https://github.com/eclipse-ee4j/servlet-api/blob/master/spec/src/main/asciidoc/servlet-spec-body.adoc#a1-changes-since-jakarta-servlet-50

markt-asf (approved):
LGTM. Happy to see this merged.
That said, there appear to be a few more deprecated methods where default implementations could be useful:

- `HttpServletResponse.encodeUrl()`
- `HttpServletResponse.encodeRedirectUrl()`
- `ServletRequest.getRealPath()`
- `HttpSession.getValue()`
- `HttpSession.getValueNames()`
- `HttpSession.putValue()`
- `HttpSession.removeValue()`

There is also the option to provide a dummy `HttpSessionContext` implementation for `HttpSession.getSessionContext()`.

markt-asf:
The deprecated methods have been removed so this PR is no longer necessary,

PR #376, opened 2020-11-27
Fix #105 - clarify requirements for path in getRealPath(String path)

### LINKED ISSUE #105: ServletContext.getRealPath(String) needs clarification
The Javadoc for getRealPath() uses the term "virtual path" but doesn't define it anywhere. Clearly, the virtual path is relative to the context root but the following are not clear:

1\. Must the path start with '/'?
2\. If the path must start with '/', what happens if it does not?
3\. If the path doesn't have to start with '/' how do you get from the provided path to a path relative to the content root?

My personal view is:
1\. Yes.
2\. Throw an IllegalArgumentException
3\. N/A

One caveat to the above is that we might allow paths of "" which are treated as equivalent to "/".

markt-asf:
I think this correctly reflects the discussion. Happy to tweak as necessary.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -427,6 +427,11 @@ public interface ServletContext {
 
     /**
      * Gets the <i>real</i> path corresponding to the given <i>virtual</i> path.
+     * 
+     * <p>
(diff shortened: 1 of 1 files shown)


gregw (approved):

martin-g (approved):

stuartwdouglas (approved):

stuartwdouglas:
It failed the formatting check, I have fixed and force pushed.

PR #387, opened 2021-01-18
Fix #377 - clarify setCharacterEncoding(null) and related

### LINKED ISSUE #377: Clarify behaviour of ServletResponse.setCharacterEncoding(null)
The Javadoc indicates that only IANA character set names are allowed but it does not define the expected behaviour for any values that are not allowed - including `null`.

Generally, I'd like to see `UnsupportedEncodingException` but that would not be a backwards compatible change so I don't see how we could do that.

The question for `null` arose from [Tomcat bug 64938](https://bz.apache.org/bugzilla/show_bug.cgi?id=64938). I can see potential use cases for allowing `null` - even if there are alternative solutions for some (all?) of those use cases.

I propose we modify the Javadoc for `ServletResponse.setCharacterEncoding(null)` along the following lines:

- Calling with `null` means behave as if no character encoding has been explicitly set
- Invalid values will be silently ignored
- The container should generate an appropriate log message if an invalid value is ignored

markt-asf:
There are two things to check here. First that the clarified behaviour is as required. There were a few edge cases where I had to make some judgement calls. Once we are happy with the defined behaviour we can polish the wording.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletResponse.java
+++ b/api/src/main/java/jakarta/servlet/ServletResponse.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
+ * Copyright (c) 1997, 2021 Oracle and/or its affiliates and others.
  * All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
@@ -20,6 +20,7 @@ package jakarta.servlet;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
 import java.util.Locale;
 
 /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/servlet/ServletResponse.java
```diff
@@ -182,17 +193,27 @@
...
      * called after the response has been committed. It does not set the response's character encoding if it is called after
      * <code>getWriter</code> has been called or after the response has been committed.
      * <p>
+     * If calling this method has an effect (as per the previous paragraph), calling this method with {@code null} clears
+     * any content type set via a previous call to this method and clears any character encoding set via a previous call to
+     * this method, {@link #setContentType} or {@link #setLocale} but does not affect any default character encoding
+     * configured via {@link ServletContext#setResponseCharacterEncoding} or the deployment descriptor.
+     * <p>
+     * Calling this method with an invalid or unrecognised character encoding has no effect until {@link #getWriter()} is
+     * called at which point a {@link UnsupportedEncodingException} will be thrown.
```
gregw, line 208: Not sure I like this.  Firstly, what if getWriter is never called and instead getOutputStream is used to write directly encoded bytes?
In that case is there an effect of the call with invalid or unrecognised character encoding?  Ie is the Content-Type header sent with the unknown charset?       If that is what we want, then we should say so explicitly here, otherwise I suggest we throw and IllegalArgumentException to be thrown in this method.

I think I lean towards allowing unknown encodings and only throwing if getWriter is called.  Perhaps it should say:
```suggestion
     * If this method is called with an unrecognised character encoding, then a subsequent call to {@link #getWriter()}
     * will throw an  {@link UnsupportedEncodingException} will be thrown.  Content for an unknown encoding can be sent 
     * with the <code>OutputStream</code> returned from {@link #getOutputStream}
```

However, perhaps we should throw `IllegalArgumentException` if the content type is badly formed (eg unclosed quotes or illegal characters).

gregw (changes requested):
Basic direction and style is good.  Just some clarification of some wording and edge cases needed.

api/src/main/java/jakarta/servlet/ServletResponse.java
markt-asf: I like the new text. I'll update the PR.

We had a user of Tomcat that needed to set an invalid content-type header due to a broken client. With that in mind I'd prefer to leave the current behaviour as is.

api/src/main/java/jakarta/servlet/ServletResponse.java
```diff
@@ -309,13 +330,17 @@
      * {@link #setContentType} has been called with a charset specification, after {@link #setCharacterEncoding} has been
      * called, after <code>getWriter</code> has been called, or after the response has been committed.
      * <p>
+     * If calling this method has an effect on the locale (as per the previous paragraph), calling this method with
+     * {@code null} clears any locale set via a previous call to this method. If calling this method has an effect on the
+     * character encoding, calling this method with {@code null} clears the previously set character encoding.
```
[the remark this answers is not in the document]
markt-asf: I prefer `clear` as I think it is more obvious that the end result is that no value is set. `reset` could be read as reverting to some previous value.

gregw (approved):

PR #391, opened 2021-02-04
Fixes # 383 Fix Javadoc for getMajorVersion and getMinorVersion

markt-asf:
The version info in the Javadoc wasn't updated for 5.0

Signed-off-by: Mark Thomas <markt@apache.org>

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
+ * Copyright (c) 1997, 2021 Oracle and/or its affiliates and others.
  * All rights reserved.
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -112,15 +112,15 @@
 
     /**
      * Returns the major version of Jakarta Servlet that this container supports. All implementations that comply with
-     * Version 4.0 must have this method return the integer 4.
+     * Version 5.0 must have this method return the integer 5.
      *
-     * @return 4
+     * @return 5
```
gregw, line 117: Why don't we just remove the hard coded value so we don't forget this again for servlet 6.0?

markt-asf: How about this?

stuartwdouglas (approved):

PR #397, opened 2021-04-15
Add implementations for Cookie.equals, Cookie.toString, and Cookie.ha…

ChristopherSchultz:
…shcode

This is good practice for any "real" class, and there are some practical reasons specifically for `equals` and `toString`.

When writing a unit-test for some servlet-related code, I was surprised to find out that two Cookie objects were not `equals()` each other even though their fields were identical. Of course, I first had to subclass `Cookie` and then override `toString` to find out that they actually *were* equal, and then discovered that `Cookie.equals` hadn't been implemented, relying on the implementation of `Object.equals` which performs a reference-equality check *only*. Only after I changed my code over to use this Cookie subclass was I able to demonstrate via unit-test that my code was working.

This was *possible* in my case because I was controlling both the input and output Cookie objects. But in some other case, one might expect the servlet container to create the new Cookie object(s) and be unable to write a (sane) test that passed due to these missing methods.

I therefore request the inclusion of these methods into the Cookie class.

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -429,7 +429,7 @@ public class Cookie implements Cloneable, Serializable {
      */
     @Override
     public int hashCode() {
-        return Objects.hash(getName(), getValue(), getVersion(), getPath(), getDomain());
+        return Objects.hash(getName(), getValue(), getVersion(), getPath(), null == getDomain() ? null : getDomain().toLowerCase());
     }
 
     /**
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -421,4 +421,86 @@ public class Cookie implements Cloneable, Serializable {
     public boolean isHttpOnly() {
         return isHttpOnly;
     }
+
+    /**
+     * Return a hash code for this Cookie.
+     *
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
I'm not sure how to get my ECA to cover Eclipse Specification Projects. I'm happy to sign any additional agreements to necessary to contribute.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -421,4 +421,86 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+            return false;
+        }
+
+        Cookie c = (Cookie)o;
+
+        return Objects.equals(getName(), c.getName())
+                && Objects.equals(getValue(), c.getValue())
+                && getVersion() == c.getVersion()
+                && Objects.equals(getPath(), c.getPath())
+                && nullSafeEqualsIgnoreCase(getDomain(), c.getDomain())
```
BalusC, line 503: This is not reflexive with hashCode.

ChristopherSchultz: 🤔

Both `hashCode` and `equals` vary by the name, value, version, path, and domain.

Aah. I see: `equals` performs case-insensitive comparison, but `hashCode` does not normalize case.

I'll provide a fix.

ChristopherSchultz:
Oops, I didn't realize that renaming a branch would close this PR. I've submitted another PR.

PR #398, opened 2021-04-18
Add implementations for Cookie's equals, toString, and hashCode

ChristopherSchultz:
This is good practice for any "real" class, and there are some practical reasons specifically for equals and toString.

When writing a unit-test for some servlet-related code, I was surprised to find out that two Cookie objects were not equals() each other even though their fields were identical. Of course, I first had to subclass Cookie and then override toString to find out that they actually were equal, and then discovered that Cookie.equals hadn't been implemented, relying on the implementation of Object.equals which performs a reference-equality check only. Only after I changed my code over to use this Cookie subclass was I able to demonstrate via unit-test that my code was working.

This was possible in my case because I was controlling both the input and output Cookie objects. But in some other case, one might expect the servlet container to create the new Cookie object(s) and be unable to write a (sane) test that passed due to these missing methods.

I therefore request the inclusion of these methods into the Cookie class.

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -429,7 +429,7 @@ public class Cookie implements Cloneable, Serializable {
      */
     @Override
     public int hashCode() {
-        return Objects.hash(getName(), getValue(), getVersion(), getPath(), getDomain());
+        return Objects.hash(getName(), getValue(), getVersion(), getPath(), null == getDomain() ? null : getDomain().toLowerCase());
     }
 
     /**
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -421,4 +421,86 @@ public class Cookie implements Cloneable, Serializable {
     public boolean isHttpOnly() {
         return isHttpOnly;
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


pzygielo:
It's a good practice to run build locally before pushing changes.
Once formatting is fixed, unknown `Objects` will be reported.

> When writing a unit-test for some servlet-related code,

I'm surprised no UT is delivered with this new functionality here.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -421,4 +421,86 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+    public boolean equals(Object o) {
+        if(null == o) {
+            return false;
+        }
+
+        if(super.equals(o)) {
+            return true;
+        }
+
+        if(!(getClass().equals(o.getClass()))) {
```
pzygielo, line 493: Why? Can't the instance of subclass be equal to given `Cookie`?
For `ArrayList` or `Hashtable` such case is not blocked, why should `Cookie` be different?

BalusC:
This is likely going to be superseded by https://github.com/eclipse-ee4j/servlet-api/pull/401

ChristopherSchultz:
> I'm surprised no UT is delivered with this new functionality here.

Touché.

ChristopherSchultz:
> This is likely going to be superseded by #401

I'm happy to apply my efforts there.

Issue #175, opened 2017-05-02
SameSite Cookie Support


glassfishrobot:
Good morning,

I'm not sure if this is the best avenue to communicate this feedback, but while working on lift/framework#1828 we discovered that the current version of the Servlet API doesn't support the [same-site cookie attribute](https://www.owasp.org/index.php/SameSite).

I would like to propose an addition to `javax.servlet.http.Cookie` to add support for this attribute in Servlet API 4.0, which would give us a path forward in the Lift Framework for supporting it.

bes:
Hi!

This issue is quite important for web security, but many of the links in this thread are dead and there seems to be little or no movement for a while on this issue.

I have tried searching the internet for more information about the Servlet API and work on SameSite but this is the best thread I could find.

Could anyone with knowledge jump in and clarify what release this could potentially end up in, or at least what the status is?

Thanks!

aldaris:
Are PRs welcome for this RFE?

markt-asf:
Yes, but...
1. We don't yet have the text for the spec document and I'd expect that would need updates as well as the code.
2. While the Servlet spec document references RFC 6265, the Javadoc references the Netscape Cookie spec and RFC 2109. I think the Javadoc needs some updates to better align it with RFC 6265  (and that might trigger some debates about backwards compatibility)
3. I don't know if anything in 2) would impact on the changes required for this issue.

Overall having a PR is better than not having one, I'm just trying to set the expectation that it might not get merged quickly.

aldaris:
Not sure if support for SameSite cookies have to be covered at the same time as when the JavaDocs need to get restructured. I have filed #271 just in case.

joakime:
Google Chrome is going to start enforcing the `SameSite` Cookie attribute.

> Chrome plans to implement the new model with Chrome 80 in February 2020. Mozilla and Microsoft have also indicated intent to implement the new model in Firefox and Edge, on their own timelines.

https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

cleankod:
Some browsers (for example Google Chrome) announced that they would require the `SameSite` directive to be present as a Cookie attribute. Otherwise the Cookie will not be sent by the browser. This was supposed to be required starting from April 2020, but due to COVID-19 crisis it was delayed ([source](https://www.chromium.org/updates/same-site)). This means we have a little more time to get everything in place.

[...]

PR #399, opened 2021-04-26
Add Cookie#setAttribute 

BalusC:
For discussion, see #175 and #271

While at it I also took opportunity to:
- cleanup `isToken` helper method whose comment and methodname were completely out of sync with actual implementation
- fix "rfc2019" typo in comments to "rfc2109"
- add missing check for "HttpOnly" in cookie name
- removed outdated javadoc wrt version 1 being "experimental"

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -20,8 +20,11 @@ package jakarta.servlet.http;
 
 import java.io.Serializable;
 import java.text.MessageFormat;
+import java.util.Collections;
 import java.util.Locale;
+import java.util.Map;
 import java.util.ResourceBundle;
+import java.util.TreeMap;
 
 /**
  *
@@ -56,7 +59,7 @@ import java.util.ResourceBundle;
  */
 public class Cookie implements Cloneable, Serializable {
 
-    private static final long serialVersionUID = -6454587001725327448L;
+    private static final long serialVersionUID = 4901306390139828294L;
 
     private static final String TSPECIALS;
 
@@ -93,6 +96,9 @@ public class Cookie implements Cloneable, Serializable {
     private int version = 0; // ;Version=1 ... means RFC 2109++ style
     private boolean isHttpOnly = false;
 
+    private Map<String, String> attributes = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
+    private Map<String, String> unmodifiableAttributes = Collections.unmodifiableMap(attributes);
+
     /**
      * Constructs a cookie with the specified name and value.
      *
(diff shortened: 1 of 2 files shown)


BalusC:
Some thoughts:

- perhaps add removeAttribute()
- perhaps detail how empty strings and nulls should be interpreted (expectation: empty string represents a boolean attribute and null indicates that attribute should not be set)

joakime:
Since you are touching the RFC / version details in the javadoc, do we want to mention that RFC2109 is obsolete and replaced with RFC2965 which is also now obsolete and replaced with RFC6265.
The use of "Version" in the Cookie is also increasingly unsupported in modern browser updates (most browsers will either ignore the version attribute, or the entire cookie if it contains a version attribute).

BalusC:
Candidate for a new ticket/PR indeed but not applicable in current context.

joakime:
This PR is especially timely, considering that Chrome and Firefox devs are discussing a new `Partitioned` attribute for the Cookie spec on the ietf-http-wg

* https://lists.w3.org/Archives/Public/ietf-http-wg/2021AprJun/0099.html
* https://lists.w3.org/Archives/Public/ietf-http-wg/2021AprJun/0102.html
* https://hacks.mozilla.org/2021/02/introducing-state-partitioning/
* https://github.com/privacycg/storage-access/issues/75
* https://discourse.wicg.io/t/proposal-cookies-having-independent-partitioned-state-chips/5290
* https://github.com/DCtheTall/CHIPS

pzygielo:
It looks that cloned Cookie shares `*attributes` with original. Is that as intended?

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -428,4 +441,56 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+     * 
+     * @param value the value of the cookie attribute associated with the given name, can be {@code null}
+     *
+     * @throws IllegalArgumentException if the cookie attribute name is null or empty, or if either the cookie attribute name or value
+     * contains any illegal characters (for example, a comma, space, or semicolon) or matches a token reserved for use by the cookie protocol
+     * 
+     * @since Servlet 5.1
+     */
+    public void setAttribute(String name, String value) {
+        if (name == null || name.isEmpty()) {
```
gregw, line 464: The implementation here needs to enforce the apidoc above that says: 
> This should override any predefined attribute for which already a getter/setter pair exist in the current version.

This can be done be removing all the fields and using only the map to store the attributes (and then updating the getters/setters accordingly).  Alternately, known fields can be intercepted here and call the setters directly.

Eitherway, we need for the known fields to always appear in the map or to never appear in the map, and never have different values if they appear in both.

gregw (changes requested):

api/src/main/java/jakarta/servlet/http/Cookie.java
BalusC: That won't work in long term. An good example is the earlier proposal to make `SameSite` an enum. We want the API to behave in best possible compatible way when HTTP guys at some point decide to add a new value to the enumeration.

BalusC: I've prepped another PR for this: https://github.com/eclipse-ee4j/servlet-api/pull/400

BalusC:
Closing off. This is superseded by https://github.com/eclipse-ee4j/servlet-api/pull/400

PR #400, opened 2021-05-02
Add Cookie#setAttribute, v2

BalusC:
As requested by @gregw here https://github.com/eclipse-ee4j/servlet-api/pull/399#discussion_r624276777

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/Cookie.java
+++ b/api/src/main/java/jakarta/servlet/http/Cookie.java
@@ -267,7 +267,7 @@ public class Cookie implements Cloneable, Serializable {
      * @see #getPath
      */
     public void setPath(String uri) {
-        attributes.put("path", uri);
+        attributes.put(PATH, uri);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


gregw (commented):
I like the semantics of this one more.... but slightly concerned with the efficiency of having to always create the map... basically on every request with a cookie.

This would be OK, so long as an implementation could override methods with a more efficient implementation for cookies that it knows have no special attributes.     But this version will not let the underlying attributes map not be created.

In reality, it was a mistake that this was not an interface, with readonly and read/write versions... ie there is no need to call a setter on a Cookie obtained from a request.

As this can effect trillions of requests, every nano second we can save here will save many Kgs of carbon.   So give me a day or two and I'll see if I can come up with a version that keeps the API, but can be more efficient for the most frequent cases.

gregw:
I've created #401 that lazily creates the attribute map, so it will not be instantiated on the most frequent usage of an inbound cookie

gregw:
@markt-asf @stuartwdouglas should we be creating unit tests for any API that we update/change?   If so... is there a test framework we should use?

stuartwdouglas:
I think unit tests are a good idea. My preference would be JUnit 5.

BalusC:
Closing off. This is superseded by https://github.com/eclipse-ee4j/servlet-api/pull/401

PR #401, opened 2021-05-03
Servlet api issue 175 add cookie set attribute v3

gregw:
Extends #400 with optimisation to lazily create the attribute map.
Also added hash and equality for completeness


stuartwdouglas (approved):
This LGTM, but what are we going to do for TCK tests for changes? AFAIK there is no real plan for how to handle changes that require TCK updates.

gregw:
@stuartwdouglas hopefully this should be transparent change to the TCK, so at least we wont break it and thus it doesn't specifically need a TCK change.... but then it is a change to the API so it should be tested in the TCK...

unless we add a unit test in the API project and then add a single new TCK test that runs the unit tests from the project?

stuartwdouglas:
Yea, its not that this will break existing stuff, but more that we should be testing that containers implement it properly. 

Maybe we can just wait until there is more clarity on the TCK development process and add the tests then.

gregw:
So OK to merge then?
@BalusC as the OP, are you good with this version?
@stuartwdouglas as this contains commits from both @BalusC and myself, then should I not squash before the merge?

gregw:
Ah, I just saw the comments on #400 regarding junit5.   I will add some unit tests for this now....

gregw:
@stuartwdouglas that was worthwhile!  Found a few little issues and corner cases that needed to be resolved.  Coverage is now 96%

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -369,22 +383,31 @@ public void setVersion(int v) {
     }
 
     /*
-     * Tests a string and returns true if the string counts as a reserved token in the Java language.
+     * Tests a string and returns true if the string contains a reserved token for the Set-Cookie header.
      *
```
markt-asf, line 387: The use of token here is confusing as token has a specific meaning in the HTTP spec. This method is checking for invalid characters, not tokens.

api/src/main/java/jakarta/servlet/http/LocalStrings.properties
```diff
@@ -19,8 +19,10 @@
 # Default localized string information
 # Localized for Locale en_US
 
-err.cookie_name_is_token=Cookie name \"{0}\" is a reserved token
+err.cookie_name_is_token=Cookie name \"{0}\" contains a reserved token
 err.cookie_name_blank=Cookie name must not be null or empty
```
markt-asf, line 23: I think this should be:
`Cookie name \"{0}"\" is a reserved token or contains an invalid character for a cookie name`

api/src/main/java/jakarta/servlet/http/LocalStrings.properties
```diff
@@ -19,8 +19,10 @@
...
+err.cookie_attribute_name_is_token=Cookie attribute name \"{0}\" contains a reserved token
```
markt-asf, line 24: I think this should be:
`Cookie attribute name \"{0}\" contains an invalid character for an attribute name.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -234,9 +247,13 @@ public void setMaxAge(int expiry) {
...
      * @see #setMaxAge
+     * 
+     * @throws NumberFormatException when this attribute is set via {@link #setAttribute(String, String)} with a value which
+     * is not in number format
      */
     public int getMaxAge() {
-        return maxAge;
+        String maxAge = getAttribute(MAX_AGE);
+        return maxAge != null ? Integer.parseInt(maxAge) : -1;
```
markt-asf, line 256: Maybe invert this for clarity?

markt-asf (changes requested):
I like it. A few minor comments/queries.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -83,22 +94,23 @@ public Boolean run() {
...
+    private Map<String, String> attributes = null;
+
+    private void putAttribute(String name, String value) {
+        if (attributes == null)
+            attributes = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        if (value == null)
+            attributes.remove(name);
+        else
+            attributes.put(name, value);
+    }
```
BalusC, line 113: Braces missing. Also strange place for this method.

api/src/main/java/jakarta/servlet/http/Cookie.java
BalusC: Please propose. Original javadoc also didn't make any sense wrt actual impl.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -426,6 +454,88 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+        if (name == null || name.isEmpty()) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_blank"));
+        }
+
+        if (containsReservedToken(name)) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_is_token", name));
+        }
+
+        if (MAX_AGE.equalsIgnoreCase(name) && value != null)
+            Long.parseLong(value);
```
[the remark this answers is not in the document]
BalusC: This is not good for forward compatibility.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -426,6 +454,88 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+     * Returns an unmodifiable mapping of all cookie attributes set via {@link #setAttribute(String, String)} as well as any
+     * predefined setter method, except for <code>version</code>.
+     * 
+     * @return an unmodifiable mapping of all cookie attributes set via <code>setAttribute(String, String)</code> as well as
+     * any predefined setter method, except for <code>version</code>
+     * 
+     * @since Servlet 5.1
+     */
+    public Map<String, String> getAttributes() {
+        return Collections.unmodifiableMap(attributes == null ? Collections.emptyMap() : attributes);
```
BalusC, line 517: emptyMap is already unmodifiable.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -83,22 +94,23 @@ public Boolean run() {
     //
     // The value of the cookie itself.
     //
-
-    private String name; // NAME= ... "$Name" style is reserved
+    private final String name; // NAME= ... "$Name" style is reserved
     private String value; // value of NAME
```
BalusC, line 98: Why is only name changed final but value not?

pzygielo:
Add something like
```diff
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.2</version>
+            </plugin>
```
to api/pom.xml/build to pick newer surefire and have tests executed (currently - https://github.com/eclipse-ee4j/servlet-api/pull/401/checks?check_run_id=2497535988#step:4:1359:
```

-------------------------------------------------------
 T E S T S
-------------------------------------------------------

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

api/src/main/java/jakarta/servlet/http/Cookie.java
gregw: Because there is a setter for the value, but not for the name. Thus only the name is immutable.

api/src/main/java/jakarta/servlet/http/Cookie.java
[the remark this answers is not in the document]
gregw: I don't think I can just remove it, as then we have to deal with what does getMaxAge() do if the value is not a number.   Either we throw here or we throw later or we treat not a number as -1.

My thought is that allowing illegal values will just make our life more difficult in future.  So if we throw, then we should throw on the setter not the getter (else we are victim blaming!).

api/src/main/java/jakarta/servlet/http/Cookie.java
gregw: hows that?

api/src/main/java/jakarta/servlet/http/LocalStrings.properties
gregw: is the property name write? 
Also the translation files are very out of date!

api/src/main/java/jakarta/servlet/http/Cookie.java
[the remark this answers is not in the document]
BalusC: Imagine we end up with 10 new non-string attributes 10 year later. Should setAttribute() have possibly 10 different fallbacks/checks for all of them and have them all documented in its javadoc? This is just code smell.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -426,6 +445,100 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+        if (MAX_AGE.equalsIgnoreCase(name) && value != null) {
+            Long.parseLong(value);
+        }
+
+        putAttribute(name, value);
+    }
+
+    private void putAttribute(String name, String value) {
+        if (attributes == null)
+            attributes = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
BalusC, line 486: Braces missing.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -426,6 +445,100 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+        if (name == null || name.isEmpty()) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_blank"));
+        }
+
+        if (hasReservedCharacters(name)) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_is_token", name));
+        }
+
+        if (MAX_AGE.equalsIgnoreCase(name) && value != null) {
+            Long.parseLong(value);
```
markt-asf, line 478: Use `Integer` to be consistent with `setMaxAge(int)` ?

api/src/main/java/jakarta/servlet/http/LocalStrings.properties
markt-asf: Fair point re the property name. Maybe `err.cookie_name_is_invalid`? Same for attribute name below.

markt-asf (commented):
LGTM. This is a good addition to the API.

api/src/main/java/jakarta/servlet/http/Cookie.java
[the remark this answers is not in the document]
gregw: I think yes.  Sure it is a PITA, but it is far better to check arguments than to allow illegal values into the map.     I have changed so that rather than duplicate the checking, it now just calls `setMaxAge`

api/src/main/java/jakarta/servlet/http/Cookie.java
gregw: I now just call `setMaxAge` here, rather than duplicate the checking logic.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -426,6 +445,102 @@ public void setHttpOnly(boolean isHttpOnly) {
...
+        if (name == null || name.isEmpty()) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_blank"));
+        }
+
+        if (hasReservedCharacters(name)) {
+            throw new IllegalArgumentException(createErrorMessage("err.cookie_attribute_name_invalid", name));
+        }
+
+        if (MAX_AGE.equalsIgnoreCase(name) && value != null) {
+            setMaxAge(Integer.parseInt(value));
```
BalusC, line 478: Javadoc needs to be updated to reflect this behavior.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -234,9 +238,13 @@ public void setMaxAge(int expiry) {
      * browser shutdown
      *
      * @see #setMaxAge
+     * 
+     * @throws NumberFormatException when this attribute is set via {@link #setAttribute(String, String)} with a value which
+     * is not in number format
```
BalusC, line 243: Javadoc needs to be updated to remove @throws as this is not anymore the case.

api/src/main/java/jakarta/servlet/http/Cookie.java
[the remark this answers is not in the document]
BalusC: I disagree.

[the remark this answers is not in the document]
gregw: Any other opinions on this? @markt-asf @stuartwdouglas 

I much prefer fail fast.   Otherwise the exception will be thrown from the getting inside addSetCookie and it will have to throw back to the application.   Much better to throw to the application when the Cookie is being created.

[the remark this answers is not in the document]
BalusC: Nope. The whole idea of forward compatibility of `attributes` is that you should use `cookie.getAttributes().get("Max-Age")` instead to obtain it (as well as all others). Otherwise `SameSite` is going to bite back when it's going to be implemented as an `enum` and HTTP guys add a new value.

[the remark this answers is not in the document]
gregw: But then something needs to check the value, what we risk sending invalid set-cookies.

I guess new fields unsupported by the api will be checked and thrown from addSetCookie.... Hmmm so maybe ok to do it all there.  I'm convincible... but I'd like to hear from others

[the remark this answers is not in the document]
stuartwdouglas: I am +1 on Greg's approach. We should not be allowing invalid values into the map.

api/pom.xml
```diff
@@ -85,6 +85,27 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.7.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest</artifactId>
+            <version>2.2</version>
```
stuartwdouglas, line 104: Should we move these to a <properties> section?

stuartwdouglas (approved):

markt-asf (approved):

api/pom.xml
gregw: move the versions to properties?

api/src/main/java/jakarta/servlet/http/Cookie.java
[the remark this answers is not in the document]
ChristopherSchultz: This could be implemented as a separate cookie-attribute-validator which could be applied at a particular time, either by a client or by a server, instead of placing these checks in `setAttribute` directly.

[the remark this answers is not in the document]
gregw: Sorry se this after the merge. I think calling the setter avoids duplicating the validation logic simply enough

api/src/main/java/jakarta/servlet/http/LocalStrings_fr.properties
```diff
@@ -19,12 +19,12 @@
...
+http.method_not_implemented=Le m�thode {0} n''est pas d�finie dans la RFC 2068 et n''est pas support�e par l''API Servlet
 
-http.method_get_not_supported=La m�thode HTTP GET n''est pas support�e par cette URL
-http.method_post_not_supported=La m�thode HTTP POST n''est pas support�e par cette URL
-http.method_put_not_supported=La m�thode HTTP PUT n''est pas support�e par cette URL
-http.method_delete_not_supported=La m�thode HTTP DELETE n''est pas support�e par cette URL 
+http.method_get_not_supported=La m�thode HTTP GET n''est pas support�e par cette URL
+http.method_post_not_supported=La m�thode HTTP POST n''est pas support�e par cette URL
+http.method_put_not_supported=La m�thode HTTP PUT n''est pas support�e par cette URL
+http.method_delete_not_supported=La m�thode HTTP DELETE n''est pas support�e par cette URL
```
BalusC, line 30: Please correct this and fix your editor to not anymore cause this in the future.

markt-asf: I've fixed the file.

PR #405, opened 2021-06-04
Issue 225 do head content length

### LINKED ISSUE #225: HttpServlet.doHead sets content-length

gregw:
This is an alternate to #356 to fix #225:
 + adds mock implementations and unit tests
 + fixes some actual bugs in nobody response (with regards to reset and resetBuffer)
 + notes some deficiencies in the nobody response
 + allows a container to disable the nobody response handling


gregw:
I'm not entirely convinced by the use of a marker interface on `ServletConfig`.  So open to suggestions of how else we could make that handling optional, specially if we think there may be other cases that the container impl may wish to direct the servlet-api impl?

Eitherway, the reset bugs should definitely be fixed.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -732,13 +767,19 @@ public void write(byte buf[], int offset, int len) throws IOException {
         contentLength += len;
     }
 
+    @Override
+    public void flush() throws IOException {
+        if (disableFlush.get() != Boolean.TRUE)
```
stuartwdouglas, line 772: This feels a bit yuck, because it is an identity comparison on Boolean.TRUE rather than the actual boolean comparison that it looks like. I know it does not really matter but maybe rewrite it to `Boolean.TRUE.equals(disableFlush.get())`?

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -586,6 +590,9 @@ public void service(ServletRequest req, ServletResponse res) throws ServletExcep
 
         service(request, response);
     }
+
+    public interface HeadHandledByContainer extends ServletConfig {
```
stuartwdouglas, line 594: I don't really like this, IMHO we should deprecate the NoBody stuff and stipulate that it is the containers responsibility (although I may be slightly biased as I know this will 'just work' in Undertow, not sure about other containers).

If we did this would could have a jakarta.servlet.legacy-head-handling as a SC param to re-enable the legacy behavior.

gregw: Ok, I'll give that a try...

markt-asf:
I am in general agreement that the container is better placed than the Servlet API to correctly generate a HEAD response.

If we are going to change the spec document so that the container is required to generate a HEAD response for a HEAD request then why do we need to retain the legacy approach? Could we have doHead() delegate to doGet() by default and leave it at that?

My assumption is that if a user Servlet overrides doHead() then that, like now, will always be used instead of the container (or legacy) method of generating the HEAD response.

If we do retain the legacy approach to HEAD generation, I like the idea of using the ServletConfig to enable/disable its use.

We also need to think about the specification language for this. It might be worth working on that as part of this PR in case thinking about that identifies any edge cases we haven't thought of.

gregw:
@markt-asf @stuartwdouglas  your thoughts on this one now?

stuartwdouglas (approved):
LGTM

gregw:
@markt-asf deprecated and added text in the spec.

gregw:
oops - I messed up commit message. I will edit to something better.

PR #408, opened 2021-07-27
Cleanup

markt-asf:
Just some simple clean-up I noticed while looking at various other issues.

### DIFF
--- a/api/src/main/java/jakarta/servlet/HttpMethodConstraintElement.java
+++ b/api/src/main/java/jakarta/servlet/HttpMethodConstraintElement.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates and others.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates and others.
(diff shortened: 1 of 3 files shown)


martin-g (approved):
You could use maven-javadoc-plugin's `failOnError` + `doclint` to catch more similar issues. Here is how we do it in Apache Wicket - https://github.com/apache/wicket/blob/90bbe8f00f0c3273a70fa756575b294ec1a008dc/pom.xml#L1063-L1064
It could be added to https://github.com/eclipse-ee4j/servlet-api/blob/c04d563cf5fbff2ebf343786076a86ee136a218e/api/pom.xml#L284

markt-asf:
Force push to correct format errors reported by CI.
@martin-g I'm currently just using the checks in my IDE. I'm a little wary of enabling `failOnError` due to the increasing strictness of Javadoc in later Java versions and not wanting to break the build. I think it is something worth doing but separately to this commit.

stuartwdouglas (approved):

gregw (approved):

PR #409, opened 2021-07-27
Next version is now planned to be 6.0 rather than 5.1

markt-asf:


stuartwdouglas (approved):

gregw:
I've also created a 6.0 project so we can start assigning issues to that project.

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -503,6 +503,8 @@ private void putAttribute(String name, String value) {
      *
      * @param name the name of the cookie attribute to set the value for, case insensitive
```
pzygielo:
```suggestion
     * @param name the name of the cookie attribute to get the value of, case insensitive
```

markt-asf: Tx. I've updated the PR to include this correction.

PR #413, opened 2021-07-29
Fix #411 No Object Wrapper Identity requirement for dispatch

### LINKED ISSUE #411: Deprecate object wrapper identity
The Servlet specification allows for wrappers to do more than just override existing methods! In 6.2.2, the specification says that:

> … the developer not only has the ability to override existing methods on the request and response objects, but to provide new API… 

So the example above could introduce new API to access the original user principal:

```java
public static class ForcedUserRequest extends HttpServletRequestWrapper 
{ 
    // ... getUserPrincipal & isUserInRole as above
    public Principal getOriginalUserPrincipal() 
    { 
        return super.getUserPrincipal(); 
    }
    public boolean isOriginalUserInRole(String role) 
    { 
        return super.isUserInRole(role); 
    } 
}
```
In order for targets to be able to use these new APIs then they must be able to downcast the passed request/response to the known wrapper type:
```java
@WebServlet(urlPatterns = {"/servletB/*"})
public static class ServletB extends HttpServlet
{
    @Override
    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
        throws ServletException, IOException
    {
        MyWrappedRequest myr = (MyWrappedRequest)req;
        resp.getWriter().printf("user=%s orig=%s wasAdmin=%b%n", 
            req.getUserPrincipal(), 
            myr.getOriginalUserPrincipal(),
            myr.isOriginalUserInRole("admin"));
    }
}
```
This downcast will only work if the wrapped object is passed through the container without any further wrapping, thus the specification requires “wrapper object identity”:

>… the container must ensure that the request and response object that it passes to the next entity in the filter chain, or to the target web resource if the filter was the last in the chain, is the same object that was passed into the doFilter method by the calling filter. The same requirement of wrapper object identity applies to the calls from a Servlet or a filter to RequestDispatcher.forward  or  RequestDispatcher.include, when the caller wraps the request or response objects.

This “wrapper object identity” requirement means that the container is unable to itself wrap requests and responses as they are passed to filters and servlets. This restriction has, directly and indirectly, a huge impact on the complexity, efficiency, and correctness of Servlet container implementations, all for very dubious and redundant benefits:

**Bad Software Components**
In the example of ServletB above, it is a very bad software component as it cannot be invoked simply by respecting the signature of its methods. The caller must have a priori knowledge that the passed request will be downcast and any other caller will be met with a ClassCastException. This defeats the whole point of an API specification like Servlets, which is to define good software components that can be variously assembled according to their API contracts.

**No Multiple Concerns**
It is not possible for multiple concerns to wrap request/responses. If another filter applies its own wrappers, then the downcast will fail. The requirement for “wrapper object identity” requires the application developer to have total control over all aspects of the application, which can be difficult with discovered web fragments and ServletContainerInitializers.

**Mutable Requests**
By far the biggest impact of “wrapper object identity” is that it forces requests to be mutable! Since the container is not allowed to do its own wrapping within RequestDispatcher.forward(...) then the container must make the original request object mutable so that it changes the value returned from getServletPath() to reflect the target of the dispatch.  It is this impact that has significant impacts on complexity, efficiency, and correctness:
- Mutating the underlying request makes the example implementation of isOriginalUserInRole(String) incorrect because it calls super.isUserInRole(String) whose result can be mutated if the target Servlet has a run-as configuration.  Thus this method will inadvertently return the target rather than the original role.
- There is the occasional need for a target Servlet to know details of the original request (often for debugging), but the original request can mutate so it cannot be used. Instead, an ever-growing list of Request attributes that must be set and then cleared on the original request attributes, just in case of the small chance that the target will need one of them.  A trivial forward of a request can thus require at least 12 Map operations just to make available the original state, even though it is very seldom required. Also, some aspects of the event history of a request are not recoverable from the attributes: the isUserInRolemethod; the original target of an include that does another include.
- Mutable requests cannot be safely passed to asynchronous processes, because there will be a race between the other thread call to a request method and any mutations required as the request propagates through the Servlet container (see the “Off to the Races” example below).  As a result, asynchronous applications SHOULD copy all the values from the request that they MIGHT later need…. or more often than not they don’t, and many work by good luck, but may fail if timing on the server changes.
- Using immutable objects can have significant benefits by allowing the JVM optimizer and GC to have knowledge that field values will not change.   By forcing the containers to use mutable request implementations, the specification removes the opportunity to access these benefits. Worse still, the complexity of the resulting request object makes them rather heavy weight and thus they are often recycled in object pools to save on the cost of creation. Such pooled objects used in asynchronous environments can be a recipe for disaster as asynchronous processes may reference a request object after it has been recycled into another request.

**Unnecessary**
New APIs can be passed on objects set as request attribute values that will pass through multiple other wrappers, coexist with other new APIs in attributes and do not require the core request methods to have mutable returns.

**Conclusion**
The “wrapper object identity” requirement has little utility yet significant impacts on the correctness and performance of implementations. It significantly impairs the implementation of the container for a feature that can be rendered unusable by a wrapper applied by another filter.  It should be removed from Servlet 6.0 and requests passed in by the container should be immutable.

gregw:
Fix #411 No Object Wrapper Identity requirement for dispatch

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -2688,20 +2688,26 @@ example, the developer may wish to extend the response object with
 higher level output objects than the output stream or the writer, such
 as API that allows DOM objects to be written back to the client.
 
-In order to support this style of filter the
-container must support the following requirement. When a filter invokes
-the `doFilter` method on the container’s filter chain implementation,
-the container must ensure that the request and response object that it
+In order to support this style of filter the container must support
+the following requirements:
+
+ * When a filter invokes the `doFilter` method on the container’s
+filter chain implementation, the container must ensure that
+the request and response objects that it
 passes to the next entity in the filter chain, or to the target web
 resource if the filter was the last in the chain, is the same object
 that was passed into the `doFilter` method by the calling filter.
-
-The same requirement of wrapper object
-identity applies to the calls from a servlet or a filter to
-`RequestDispatcher.forward` or `RequestDispatcher.include`, when the
-caller wraps the request or response objects. In this case, the request
-and response objects seen by the called servlet must be the same wrapper
-objects that were passed in by the calling servlet or filter.
+ * When a filter or servlet calls `RequestDispatcher.forward` or
+`RequestDispatcher.include`, then the request and response objects
+seen by the called filter(s) and/or servlet must either:
+be the same wrapper objects that were passed; or wrappers of the
+objects that were passed.
+ * When `startAsync(ServletRequest, ServletResponse)`
+is used to commence an asynchronous cycle then the request and response objects
+seen by any filter(s) and/or servlet subsequent to an `AsyncContext.dispatch()`
+(or overloaded variant) must either:
+be the same wrapper objects that were passed; or wrappers of the
+objects that were passed.
 
 ==== Filter Environment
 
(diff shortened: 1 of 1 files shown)


stuartwdouglas (approved):
I think this looks good. 

If this requirement was removed from the filter chain I would be a bit more conservative, but just removing it from RequestDispatcher and async dispatch is fine, and could greatly simplify our implementation.

PR #414, opened 2021-07-29
Fix #412 AsyncContext execute

### LINKED ISSUE #412: Request Serialized Threads
The example below is a very typical race that exists in many applications between a timeout and asynchronous processing:
```java
@Override
protected void doGet(HttpServletRequest request,
                     HttpServletResponse response) throws IOException
{
    AsyncContext async = request.startAsync();
    PrintWriter out = response.getWriter();
    async.addListener(new AsyncListener() 
    {
        @Override
        public void onTimeout(AsyncEvent asyncEvent) throws IOException
        {
            response.setStatus(HttpServletResponse.SC_BAD_GATEWAY);
            out.printf("Request %s timed out!%n", request.getServletPath());
            out.printf("timeout=%dms%n ", async.getTimeout());
            async.complete();
        }
    });

    CompletableFuture<String> logic = someBusinessLogic();
    logic.thenAccept(answer ->
    {
        response.setStatus(HttpServletResponse.SC_OK);
        out.printf("Request %s handled OK%n", request.getServletPath());
        out.printf("The answer is %s%n", answer);
        async.complete();
    });
}
```
Because the handling of the result of the business logic may be executed by a non-container-managed thread, it may run concurrently with the timeout callback. The result can be an incorrect status code and/or the response content being interleaved. Even if both lambdas grab a lock to mutually exclude each other, the results are sub-optimal, as both will eventually execute and one will ultimately throw an IllegalStateException, causing extra processing and a spurious exception that may confuse developers/deployers.

The current specification of the asynchronous life cycle is the worst of both worlds for the implementation of the container. On one hand, they must implement the complexity of request-serialized events, so that for a given request there can only be a single container-managed thread in service(...), doFilter(...), onWritePossible(), onDataAvailable(), onAllDataRead()and onError(), yet on the other hand an arbitrary application thread is permitted to concurrently call the API, thus requiring additional thread-safety complexity. All the benefits of request-serialized threads are lost by the ability of arbitrary other threads to call the Servlet APIs.

**Request Serialized Threads**

The fix is twofold: firstly make more Servlet APIs immutable (as discussed #411) so they are safe to call from other threads;  secondly and most importantly, any API that does mutate state should only be able to be called from request-serialized threads!   The latter might seem a bit draconian as it will make the lambda passed to thenAccept in the example above throw an `IllegalStateException` when it tries to `setStatus(int)` or call `complete()`, however, there are huge benefits in complexity and correctness and only some simple changes are needed to rework existing code.

Any code running within a call to `service(...)`, `doFilter(...)`, `onWritePossible()`, `onDataAvailable()`, `onAllDataRead()` and `onError()` will already be in a request-serialized thread, and thus will require no change. It is only code executed by threads managed by other asynchronous components (e.g. the lambda passed to thenAccept() above) that need to be scoped. There is already the method `AsyncContext.start(Runnable)` that allows a non-container thread to access the context (i.e. classloader) associated with the request. An additional similar method `AsyncContext.dispatch(Runnable)` can be provided that not only scopes the execution but mutually excludes it and serializes it against any call to the methods listed above and any other dispatched Runnable. The Runnables passed may be executed within the scope of the dispatch call if possible (making the thread momentarily managed by the container and request serialized) or scheduled for later execution.  Thus calls to mutate the state of a request can only be made from threads that are serialized.

To make accessing the `dispatch(Runnable)` method more convenient, an executor can be provided with `AsyncContext.getExecutor()` which provides the same semantic.  The example above can now be simply updated:
```java
@Override
protected void doGet(HttpServletRequest request,
                     HttpServletResponse response) throws IOException
{
    AsyncContext async = request.startAsync();
    PrintWriter out = response.getWriter();
    async.addListener(new AsyncListener() 
    {
        @Override
        public void onTimeout(AsyncEvent asyncEvent) throws IOException
        {
            response.setStatus(HttpServletResponse.SC_BAD_GATEWAY);
            out.printf("Request timed out after %dms%n ", async.getTimeout());
            async.complete();
        }
    });

    CompletableFuture<String> logic = someBusinessLogic();
    logic.thenAcceptAsync(answer ->
    {
        response.setStatus(HttpServletResponse.SC_OK);
        out.printf("The answer is %s%n", answer);
        async.complete();
    }, async.getExecutor());
}
```
Because the `AsyncContext.getExecutor()` is used to invoke the business logic consumer, then the timeout and business logic response methods are mutually excluded. Moreover, because they are serialized by the container, the request state can be checked between each, so that if the business logic has completed the request, then the timeout callback will never be called, even if the underlying timer expires while the response is being generated. Conversely, if the business logic result is generated after the timeout, then the lambda to generate the response will never be called.  Because both of the tasks in this example call complete, then only one of them will ever be executed.

Note that the new APIs could be introduced in 6.0 without preventing the state APIs being called from other threads (perhaps they just give a warning the first time they are so called).  The prevention of calling the APIs from arbitrary threads can be done in a 6.1 or 7.0 version of the spec.

gregw:
Fix #412 AsyncContext is an Executor that mutually excludes from other container invocations for the same request.

### DIFF
--- a/api/src/main/java/jakarta/servlet/AsyncContext.java
+++ b/api/src/main/java/jakarta/servlet/AsyncContext.java
@@ -17,6 +17,9 @@
 
 package jakarta.servlet;
 
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.Executor;
+
 /**
  * Class representing the execution context for an asynchronous operation that was initiated on a ServletRequest.
  *
@@ -36,9 +39,17 @@ package jakarta.servlet;
  * {@link #dispatch} methods, call {@link #complete}.</li>
  * </ol>
  *
+ * <p>
+ * An AsyncContext is an {@link Executor} that provides mutual exclusion between container managed threads for the
+ * same request.  Specifically container managed threads are used for calls to
+ * {@link Servlet#service(ServletRequest, ServletResponse)},
+ * {@link Filter#doFilter(ServletRequest, ServletResponse, FilterChain)},
+ * {@link ReadListener#onDataAvailable()}, {@link WriteListener#onWritePossible()}
+ * and to call the {@link Runnable}s passed to {@link AsyncContext#execute(Runnable)} and {@link AsyncContext#run(Runnable)}.
+ *
  * @since Servlet 3.0
  */
-public interface AsyncContext {
+public interface AsyncContext extends Executor {
 
     /**
      * The name of the request attribute under which the original request URI is made available to the target of a
@@ -188,6 +199,11 @@ public interface AsyncContext {
      * within the same asynchronous cycle will result in an IllegalStateException. If startAsync is subsequently called on
      * the dispatched request, then any of the dispatch or {@link #complete} methods may be called.
      *
+     * <p>
+     * This method should be called from a container managed thread associated with the same request.
+     * A container may log a warning the first time this method is called from a non container managed thread and
+     * future versions of the specification may prohibit such calls.
+     *
      * @throws IllegalStateException if one of the dispatch methods has been called and the startAsync method has not been
      * called during the resulting dispatch, or if {@link #complete} was called
      *
@@ -217,6 +233,11 @@ public interface AsyncContext {
      *
      * <p>
      * See {@link #dispatch()} for additional details, including error handling.
+     *
+     * <p>
+     * This method should be called from a container managed thread associated with the same request.
+     * A container may log a warning the first time this method is called from a non container managed thread and
+     * future versions of the specification may prohibit such calls.
      *
      * @param path the path of the dispatch target, scoped to the ServletContext from which this AsyncContext was
      * initialized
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
Conceptually I like it. The implementation is not super simple if you have IO threads in the mix though. For performance reasons you will likely want to execute `ReadListener#onDataAvailable()` and `WriteListener#onWritePossible()` on an IO thread, so if you just use a simple lock for mutual exclusion you can potentially block the IO thread (which is a big no-no). This can obviously be worked around. 

I am also not really sure about `run()` and its possibilities for deadlocks. execute feels a lot safer, I can't really think of a scenario where run makes sense (also plenty of possibility for confusion between the two methods).

gregw:
@stuartwdouglas `onDataAvailable` and `onWritePossible` are already mutually excluded from other container managed threads (eg lingering in `service` or during a `dispatch`) and from each other, so the potential for blocking is already there.   So I will update the text on all these methods to make it clear that they should not be long running... or else.   As for implementation, I expect we all have some mechanism already for mutually excluding the container managed threads, so the intention here is to just slightly complicate that an let an arbitrary task be similarly excluded.

The idea is that in future, such tasks would be the preferred (only?) way of doing a `complete`, but for now continuing to allow the wild west of anybody can call at any time is necessary (as it is too late to really validate a change).   I'll rework the wording for that as well to make it clear that this is not changing existing behavior for this release, but that the behavior may change in a future release.

I'll ponder `run`... if I can't find a good use-case, then I'll remove.  Execute allows for the task to be executed in the calling context, so it can avoid a thread handoff, yet never block, so it is probably sufficient.

stand by....

PR #419, opened 2021-09-20
Remove deprecated code

markt-asf:
If this is acceptable then changes will be required to the specification document as as well. If there is consensus to apply this PR then I'll add another commit to update the specification document (I haven't done that yet in case there isn't agreement to do this).


api/src/main/java/jakarta/servlet/http/HttpUtils.java
```diff
@@ -1,249 +0,0 @@
...
-     * The keys and values in the hashtable are stored in their decoded form, so any + characters are converted to spaces,
-     * and characters sent in hexadecimal notation (like <i>%xx</i>) are converted to ASCII characters.
-     *
-     * @param s a string containing the query to be parsed
-     *
-     * @return a <code>HashTable</code> object built from the parsed key-value pairs
-     *
-     * @exception IllegalArgumentException if the query string is invalid
-     */
-    public static Hashtable<String, String[]> parseQueryString(String s) {
```
pzygielo, line 63: @markt-asf Could you suggest replacement for this method, please?

markt-asf:
Note to self. Need to update copyright years too.

gregw (approved):
Brutal but long overdue.

api/src/main/java/jakarta/servlet/http/HttpUtils.java
markt-asf: `ServletRequest.getParameterMap()`. If an application needs to differentiate between query string parameters and request body parameters it will need to implement code to do that itself. I suspect parsing the query string would be the simplest to implement.

pzygielo: @markt-asf - thank you for confirmation. I was exactly thinking about body params being included in `getParameterMap`. It's clear now.

stuartwdouglas (approved):

Issue #18, opened 2011-10-07
Clarify behaviour of getRequestURI(), getContextPath(), getServletPath() and getPathInfo()


glassfishrobot:
The specification is unclear for some or all of the above methods how the following should be handled.

*   path parameters - should they be included in the returned values or not?
*   normalisation - should the returned values be pre- or -post any normalisation?
*   url decoding - should the returned values be decoded or include the escapes?
*   character encoding - if a value is decoded, what character encoding should be used?
*   welcome files - if a request maps to a welcome file, should it be included in the return values?

markt-asf:
An additional point requiring clarification is that RFC 3986 (section 2.2) states that %nn encoded reserved characters are not equivalent to their decoded forms. i.e. "/foo/bar/" and "/foo%2Fbar" are not equivalent. This raises an additional question. When a %nn reserved character appears in any part of the URI should it be decoded or not in the return value of each of the above methods? I'm leaning towards a solution that leaves them in their %nn form and adds a utility method to the API that performs %nn decoding.

gregw:
the servlet spec in section 3.1 says: 
> Path parameters that are part of a GET request (as defined by HTTP/1.1) are not exposed by these APIs. They must be parsed from the String values returned by the getRequestURI method or the getPathInfo method.

I think when those words were written URIs were defined by [RFC2396](https://tools.ietf.org/html/rfc2396#section-3.3) which well defined path parameters in section 3.3:
```abnf
      path          = [ abs_path | opaque_part ]
      path_segments = segment *( "/" segment )
      segment       = *pchar *( ";" param )
      param         = *pchar
      pchar         = unreserved | escaped |
                      ":" | "@" | "&" | "=" | "+" | "$" | ","
```

However, that RFC has since been obsoleted by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.3) which now no longer has a normative definition for path parameters and instead just says:

> Aside from dot-segments in hierarchical paths, a path segment is
   considered opaque by the generic syntax.  URI producing applications
   often use the reserved characters allowed in a segment to delimit
   scheme-specific or dereference-handler-specific subcomponents.  For
   example, the semicolon (";") and equals ("=") reserved characters are
   often used to delimit parameters and parameter values applicable to
   that segment.  The comma (",") reserved character is often used for
   similar purposes.  For example, one URI producer might use a segment
   such as "name;v=1.1" to indicate a reference to version 1.1 of
[...]

PR #420, opened 2021-09-21
Issue #18 path modes

gregw:
For #18 
@markt-asf this is my first draft of the things we've been discussing in #18.  I think a bit of work is needed and there are still some changes that are at best to be considered as thought bubbles.
I've allowed servlet-api committers to edit this branch so update if you like, alternatively if you think this is in the ballpark, then I can move it to a branch in eclipse-ee4j repo so we can work on it collaboratively.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -272,10 +272,11 @@ public interface ServletContext {
     public URL getResource(String path) throws MalformedURLException;
 
     /**
-     * TODO
+     * TODO Unlike {@link #getPath(String)}, returns only existing resources.
      * 
      * @param path a <code>String</code> specifying the path to the resource
-     * @return Path to the resource... which may be a temporary file extracted from a jar.
+     * @return Path to the resource or null if it does not exist. Path may be to real file, a temporary file extracted from
+     * a jar or to Path within a special {@link java.nio.file.FileSystem} such as ZIP.
      * @throws MalformedURLException
      */
     Path getResourceAsPath(String path) throws MalformedURLException;
@@ -472,10 +473,11 @@ public interface ServletContext {
 
     /**
      * TODO
-     *
+     * 
      * @param path the path relative to the context
      * @return The Path of the resource or null if a translation cannot be performed. The path may represent a non-existent
      * resource.
+     * @see #getResourceAsPath(String)
      */
     Path getPath(String path);
 
@@ -1471,8 +1473,10 @@ public interface ServletContext {
      * <li>{@link ServletContext#getContext(String)}</li>
      * <li>{@link ServletContext#getRequestDispatcher(String)}</li>
      * <li>{@link ServletContext#getRealPath(String)}</li>
+     * <li>{@link ServletContext#getPath(String)}</li>
      * <li>{@link ServletContext#getResource(String)}</li>
      * <li>{@link ServletContext#getResourceAsStream(String)}</li>
+     * <li>{@link ServletContext#getResourceAsPath(String)}</li>
      * <li>{@link ServletContext#getResourcePaths(String)}</li>
      * <li>{@link AsyncContext#dispatch(String)}</li>
      * <li>{@link AsyncContext#dispatch(ServletContext, String)}</li>
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -1494,8 +1494,8 @@ public interface ServletContext {
         /**
          * Effected methods will return/accept paths as was done prior to the introduction of the PathCodingMode in Servlet 6.0.
          * The path returned by {@link HttpServletRequest#getContextPath()} is not decoded. Paths returned by the methods
-         * {@link HttpServletRequest#getServletPath()} and {@link HttpServletRequest#getPathInfo()} are fully decoded and all
-         * path parameters removed. Decoding may include ambiguous characters such as a decoded <code>%2F</code>, which is
+         * {@link HttpServletRequest#getServletPath()} and {@link HttpServletRequest#getPathInfo()} are fully decoded and have
+         * all path parameters removed. Decoding may include ambiguous characters such as a decoded <code>%2F</code>, which is
          * indistinguishable from a segment separator '/' in the request URI. Methods accepting paths may accept encoded
          * characters. A '?' character in a path will be treated as indicating a query string and will not be encoded. Some
          * implementation specific behavioural differences may be apparent in this mode.
@@ -1516,8 +1516,8 @@ public interface ServletContext {
          * decode all non-reserved characters and remove only the <code>jsessionid</code> path parameter. All other path
          * parameters will be returned and considered part of the path for the purposes of path matching. Methods that take
          * paths expect the reserved characters to be encoded if they are to be considered part of a path segment, otherwise
-         * they will not be encoded and will have their reserved interpretation if possible. If a valid interpretation of a non
-         * encoded reserved character is not possible, then the method will throw {@link MalformedURLException} if declared,
+         * they will not be encoded and will have their reserved interpretation if possible. If a valid interpretation of a
+         * non-encoded reserved character is not possible, then the method will throw {@link MalformedURLException} if declared,
          * otherwise a {@link IllegalArgumentException}.
          */
         ENCODED,
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -1571,9 +1571,15 @@ public interface ServletContext {
         /**
          * Effected methods will only return/accept paths that do not contain any encoded reserved characters and with all path
          * parameters removed. If the requested URI contains encoded reserved characters, then a call to a method that returns
-         * such a path will throw {@link IllegalStateException}. If a path with encoded reserve characters is passed to a method
-         * taking a path as a parameter, then it will throw {@link MalformedURLException} if declared. otherwise a
-         * {@link IllegalArgumentException}.
+         * such a path will throw {@link IllegalStateException}.
+         *
+         * <p>
+         * TODO pick between this: If a path with encoded reserve characters is passed to a method taking a path as a parameter,
+         * then it will throw {@link MalformedURLException} if declared. otherwise a {@link IllegalArgumentException}.
+         *
+         * <p>
+         * TODO or this: If a path with encoded reserve characters is received, then the request will be rejected by the servlet
+         * container with a 400 BAD REQUEST response.
          */
         DECODED,
 
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -275,21 +275,20 @@ public interface ServletContext {
      * Returns a Path to the resource that is mapped to the given path String.
      *
      * <p>
-     * The path String must begin with a <tt>/</tt> and is interpreted as relative to the
-     * current context root, or relative to the <tt>/META-INF/resources</tt> directory of
-     * a JAR file inside the web application's <tt>/WEB-INF/lib</tt> directory.
-     * This method will first search the document root of the web application for the requested
-     * resource, before searching any of the JAR files inside <tt>/WEB-INF/lib</tt>.
-     * The order in which the JAR files inside <tt>/WEB-INF/lib</tt> are searched is undefined.
+     * The path String must begin with a <tt>/</tt> and is interpreted as relative to the current context root, or relative
+     * to the <tt>/META-INF/resources</tt> directory of a JAR file inside the web application's <tt>/WEB-INF/lib</tt>
+     * directory. This method will first search the document root of the web application for the requested resource, before
+     * searching any of the JAR files inside <tt>/WEB-INF/lib</tt>. The order in which the JAR files inside
+     * <tt>/WEB-INF/lib</tt> are searched is undefined.
      *
      * <p>
-     * This method allows the servlet container to make a resource available to servlets from any source.
-     * Resources can be located on a local or remote file system, in a database, or in a <code>.war</code> file.
-     * Resources may also be copied/extracted to a temporary or alternative location.
+     * This method allows the servlet container to make a resource available to servlets from any source. Resources can be
+     * located on a local or remote file system, in a database, or in a <code>.war</code> file. Resources may also be
+     * copied/extracted to a temporary or alternative location.
      *
      * <p>
-     * The servlet container or JVM must provide the {@link java.nio.file.FileSystem}s that are necessary to
-     * access the resource.
+     * The servlet container or JVM must provide the {@link java.nio.file.FileSystem}s that are necessary to access the
+     * resource.
      *
      * <p>
      * This method returns <code>null</code> if no resource is mapped to the pathname.
@@ -302,12 +301,12 @@ public interface ServletContext {
      * source code. Use a <code>RequestDispatcher</code> instead to include results of an execution.
      *
      * <p>
-     * This method has a different purpose than <code>java.lang.Class.getResource</code>,
-     * which looks up resources based on a class loader. This method does not use class loaders.
+     * This method has a different purpose than <code>java.lang.Class.getResource</code>, which looks up resources based on
+     * a class loader. This method does not use class loaders.
      *
      * <p>
-     * This method has a different purpose than {@link #getRealPath(String)} and {@link #getPath(String)},
-     * which can return non-existent resource.
+     * This method has a different purpose than {@link #getRealPath(String)} and {@link #getPath(String)}, which can return
+     * non-existent resource.
      *
      * <p>
      * This method bypasses both implicit (no direct access to WEB-INF or META-INF) and explicit (defined by the web
@@ -315,8 +314,7 @@ public interface ServletContext {
      * provided data) and when using the result not to create a security vulnerability in the application.
      * 
      * @param path a <code>String</code> specifying the path to the resource
-     * @return Path to the resource or null if it does not exist.
-     * Path may be to real file, a temporary file extracted from
+     * @return Path to the resource or null if it does not exist. Path may be to real file, a temporary file extracted from
      * a jar or to Path within a special {@link java.nio.file.FileSystem} such as ZIP.
      * @throws MalformedURLException if the path is not given in the correct form.
      * @see #getPath(String)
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1459,87 @@
...
+     * for the purposes of these modes:
+     * <ul>
+     * <li>a 'reserved character' is a character defined as <code>reserved</code> by
+     * <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-2.2">RFC3986 Section 2.2</a> or the ';'
+     * character.</li>
+     * <li>a 'path parameter' is the part of a URI segment as defined by the obsoleted
+     * <a href="https://datatracker.ietf.org/doc/html/rfc2396#section-3.3">RFC2396 Section 3.3</a>.</li>
+     * </ul>
+     */
+    public enum PathCodingMode {
```
gregw, line 1493: The approach I landed on is to have the 3 modes:   `LEGACY` is what we have now and will be the default for any web.xml <= 5.0. The `DECODED` mode is more or less what we've called `SAFE` before, it will be the default for >=6.0 and paths will not have any encodings in them and encoded reserved characters are just illegal.  `ENCODED` is the optional mode that allows encoded reserved characters.

I might be persuaded that the 6.0 default should be `ENCODED`... in which case perhaps we could drop `DECODED` and go to a boolean mode? But naming is difficult and I can't convince myself 100% of this.

I've included path parameters behaviours, but perhaps that needs a n independent mode?

We could put something in web.xml for this... but let's finalize the design before thinking about that.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1517,87 @@
...
+     * {@link #getEffectiveMajorVersion()} less that or equal to 5, otherwise it will be {@link PathCodingMode#DECODED}. The
+     * default may be changed by container specific mechanisms or overridden by a call to
+     * {@link #setPathCodingMode(PathCodingMode)}
+     *
+     * @return The path coding mode for this context.
+     */
+    PathCodingMode getPathCodingMode();
+
+    /**
+     * Set the path coding mode for the context. This method may only be called
```
stuartwdouglas, line 1595: Looks like this javadoc is incomplete

gregw: yep - forgot to `mvn install` before a push.  Work in progress, so changes and reformats coming in.... but any early thoughts appreciated to make sure I'm barking at a tree that's at least in the right forest.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1517,87 @@
...
+         * all path parameters removed. Decoding may include ambiguous characters such as a decoded <code>%2F</code>, which is
+         * indistinguishable from a segment separator '/' in the request URI. Methods accepting paths may accept encoded
+         * characters. A '?' character in a path will be treated as indicating a query string and will not be encoded. Some
+         * implementation specific behavioural differences may be apparent in this mode.
+         */
+        LEGACY,
+
+        /**
+         * Effected methods will only return/accept paths that do not contain any encoded reserved characters and with all path
+         * parameters removed. If the requested URI contains encoded reserved characters, then a call to a method that returns
```
stuartwdouglas, line 1573: Shouldn't this just return bad request and be done with it? Historically these methods (e.g. `getPathInfo`) have never thrown an exception, so changing this would potentially have unexpected consequences (especially as this can be triggered by a remote user). If an app is not using try/finally properly this could potentially result in a resource leak, and the possibility to DOS a poorly coded app by sending requests with reserved characters.

gregw: I'd be ok with this (it is essentially what Jetty already does unless you configure exceptions for strange looking URIs).

My only concern is that apps that currently do use `getRequestURI` and do their own handling will need to be modified/configured to set the `ENCODED` mode rather than just working out of the box.      But there is no way to know a-priori if an application will call the affected APIs or not, so either we delay enforcement and throw (which may have some problems) or block early with a 400 (may require some config for existing apps)

stuartwdouglas: Won't existing apps be using LEGACY anyway?

stuartwdouglas: I like the alternative.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1523,97 @@
...
+         *
+         * <p>
+         * TODO pick between this: If a path with encoded reserve characters is passed to a method taking a path as a parameter,
+         * then it will throw {@link MalformedURLException} if declared. otherwise a {@link IllegalArgumentException}.
+         *
+         * <p>
+         * TODO or this: If a path with encoded reserve characters is received, then the request will be rejected by the servlet
+         * container with a 400 BAD REQUEST response.
+         */
+        DECODED,
```
markt-asf, line 1584: I think this can be simplified. Once path parameters have been removed, it is safe to decode any %nn reserved characters apart from `%2F`. Therefore, I think we can remove the requirement to throw an error in this mode entirely. This mode then becomes:
- decode unreserved %nn
- remove all path parameters
- normalize
- decode reserved %nn except `%2F`

This achieves the same result without causing problems for apps (like Jira) that use encoded reserved characters in some requests and removes the need to triggering exceptions that will break some apps that would otherwise work correctly.

Note: There is an interesting edge case here. `/aaa/..;a=b/`. The proposal above would result in `/`. However it can be argued that this is inconsistent with RFC 3986. There are other approaches. Given there is no legitimate reason I can think of for sequences like this I quite like rejecting such requests with a 400 response.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1523,97 @@
...
+        /**
+         * Effected methods will return/accept paths in which only reserved characters are encoded. Methods returning paths will
+         * decode all non-reserved characters and remove only the <code>jsessionid</code> path parameter. All other path
+         * parameters will be returned and considered part of the path for the purposes of path matching. Methods that take
+         * paths expect the reserved characters to be encoded if they are to be considered part of a path segment, otherwise
+         * they will not be encoded and will have their reserved interpretation if possible. If a valid interpretation of a
+         * non-encoded reserved character is not possible, then the method will throw {@link MalformedURLException} if declared,
+         * otherwise a {@link IllegalArgumentException}.
+         */
+        ENCODED,
```
markt-asf, line 1595: If path parameters may be present in any segment present then it is not going to be possible write a `<url-pattern>` that matches the servlet path. The only place path parameters will be usable is in path info. I think that is too restrictive.

I currently favour a mode that maps requests to servlets based using the simplified version of `DECODED` that I described above but for methods that return path, those paths include the path parameters as described here. The mapped path (i.e. the one without the path parameters) remains available via `HttpServletMapping`.

The idea is that with this approach things are essentially unchanged for the majority of applications. The small number of applications using path parameters may need changes. Containers have the option of providing a `LEGACY` mode if there is sufficient demand from their user base from users with applications that use path parameters. I suspect that given that most apps are going to migrate from Jakarta EE 8 to Jakarta EE 10 then the few apps where changes to path parameters have an impact will recode the relevant bits of their app to handle the changes.

markt-asf (commented):
Lots of detail in the comments but the overview is I don't (currently) think we need modes:
- Make `LEGACY` a container specific option
- Simplify `DECODED` mode
- Use `DECODED` mode to map requests
- Use `ENCODED` mode for paths in the API (return values and parameters)

api/src/main/java/jakarta/servlet/ServletContext.java
gregw: I'd like to get to boolean approach too.... but so far I've been unable to reduce down the behaviours to just two.   So let's iterate on how we see the various modes working... and if we get down to 2 of them, we can go to the binary model and get rid of this enum and associated setter.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1523,97 @@
...
+     * </ul>
+     * 
+     * @since Servlet 6.0
+     */
+    public enum PathCodingMode {
+        /**
+         * Effected methods will return/accept paths as was done prior to the introduction of the PathCodingMode in Servlet 6.0.
+         * The path returned by {@link HttpServletRequest#getContextPath()} is not decoded. Paths returned by the methods
+         * {@link HttpServletRequest#getServletPath()} and {@link HttpServletRequest#getPathInfo()} are fully decoded and have
+         * all path parameters removed. Decoding may include ambiguous characters such as a decoded <code>%2F</code>, which is
```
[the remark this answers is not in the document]
gregw: I don't see how we can avoid decoding %2F as the current api says that the strings are decoded, so any container that doesn't decode is in violation.   For jetty we by default reject such requests with a 400 unless the application is configured to say that it is prepared to accept such  decoded strings.

For me that is the key difference between `LEGACY` and `DECODED` mode: the former will allow containers to act as they have previous, whilst the later is a well defined mode that never returns a decoded %2F because they have been rejected with a 400.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1523,97 @@
...
+         * indistinguishable from a segment separator '/' in the request URI. Methods accepting paths may accept encoded
+         * characters. A '?' character in a path will be treated as indicating a query string and will not be encoded. Some
```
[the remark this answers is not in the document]
gregw: I think some characters can be encoded by at least some of the methods.  For example, in LEGACAY mode  `getRequestDispatcher("/foo bar")` may result in a dispatcher for `/foo%20bar`.          But the main intention here is to say that if you have a resource called `foo?bar` then you need to pass in `/foo%3Fbar` and not `/foo?bar`

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -1436,4 +1523,97 @@
...
+         * implementation specific behavioural differences may be apparent in this mode.
+         */
+        LEGACY,
```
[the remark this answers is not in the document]
gregw: yep - very happy to get rid of modal behaviour.   Let's iterate on what we think the minimal modes are and try to reduce to 2... or better yet 1

api/src/main/java/jakarta/servlet/ServletContext.java
gregw: done

api/src/main/java/jakarta/servlet/ServletContext.java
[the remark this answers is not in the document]
stuartwdouglas: By default Undertow does not decode these, however if I had my time again I would respond with a 400 instead. Either way we will need a container specific option to allow this, as some people want it, but IMHO we should be strict about rejecting this in the spec.

Although something I just thought of is that we likely need to be clearer about reserved characters in the different parts of the URI, e.g. %2F is likely fine as part of a query string. A query string does not really have the same security issues as the actual URI, so rejecting encoded reserved characters in the query string is almost certainly the wrong thing to do.

api/src/main/java/jakarta/servlet/ServletContext.java
gregw: The beauty of what you propose here is that it get's us really close to not needing to be modal, as we would only ever return sane paths here and insane ones would be rejected with 400.  

But I'm not so sure that %2F is really the only problem character?  

Potentially there are also issues with URIs like `/aaa/..%5c../foo` which according to RFC3986 should decode to `/aaa/..\../foo`, which can be missinterpreted on some file systems.    

We've also seen issues with URIs like `/foo%00`.  Ah but this indicates a problem with specifing in terms of [reserved](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3) from the RFC.  I think we probably should specify in terms of [unreserved](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3), as there is a gap between those definitions that includes control characters.

Also if we allow encoded '%' (%25), then the resulting paths can contain percent sequences that can be dangerous to pass onto methods that a little flexible about receiving encoded characters or not.  So for example `/foo/%252E%252E/bar` could result in a `/foo/%2e%2e/bar` path that may then get later converted to `/foo/../bar` and normalized to just `/bar`

So I'm not totally convinced that %2F is the only danger encoding.

Regarding the edge case of `/aaa/..;`, jetty is now rejecting such requests with 400 unless told to do otherwise.
The question is, if we do let such requests through, should the result be `/` or `/aaa/..` ?  The later is compliant with RFC3986 but potentially dangerous when passed to file handling code.   I think your process about would result in `/`

So to be compliant with RFC3986 we'd could have a process like:
 + decode unreserved %nn
 + normalize
 + remove all path parameters
 + reject ambigous/scary/insane URIs. eg containing %2F, ".." etc.  See [Jetty's enumeration of scary URIs](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-http/src/main/java/org/eclipse/jetty/http/UriCompliance.java).
 + decode reserved %nn 

But despite these reservations, I think the general approach of path methodes never returning scary paths may be workable.   But my concern is that we then move the "modes" to the definition of what is scary.  Is that something we need to standardize? I think we need to at least enumerate the ones that should be rejected by default... we could then perhaps leave it up to containers to configure exceptions.    

However, if we want applications that handle the exceptional cases by calling getRequestURI to be portable, then maybe we need to only fail scary URIs if the path methods are called rather than reject up front?

I also think we need to think through the modes on how methods that accept paths act.  Currently I think there is a lot of flexibility, which is why so many scary URIs can be damaging if let flow through to these methods.    I'm not sure how we can make them stricter without introducing at least a LEGACY mode.

Geeze this is a horrid problem!

api/src/main/java/jakarta/servlet/ServletContext.java
gregw: I don't like treating ';' as a normal character for path methods, but special for path matching.    I think it needs to be one or the other, not both... specially when we consider the complexities of methods taking paths.

We could continue to treat ';' as special and remove all parameters from path methods. Apps wanting to handle them would need to use the requestURI as they do now.   But we need to specify handling of scary URIs that use ';' to hide '..' etc.

Or we treat ';' as a character allowed in a segment (as is now apparently is). Then `/..;` is not a problem because the segment is `..;` and not `..`.   But if we are treating is as part of a segment, the it needs to be able to be mapped.   I don't see it is impossible to map that because ';' can be included in a `<url-pattern>` as the spec says that all other parts of a pattern string are used for exact matches, so they can contain `;`, `\`, `?` and other scary characters.

gregw:
@markt-asf I've kind of rebutted all your comments..... but I don't  reject your basic premise... my comments are really just me thinking out loud.   I do think there is potential value in the general approach you are advocating - ie make the returns from the path methods non-modal and handle the scary URIs in a different way.      My concern remains defining scary URIs and then making sure we equally consider the risks of methods taking path strings.

This non modal approach is essentially how jetty is working now... at least for the methods producing paths.   

I'm going to ponder your feedback more as getting rid of modes would be good and I don't think they are perfect either.   But the good aspect of modes is that they do allow us to be stricter on what methods accepting paths will accept.

markt-asf (commented):
@gregw Thanks for your comments re rebuttal. I feel much the same way but couldn't figure out how to express it.

I understand and share your concerns regarding security. Tomcat has had to deal with a bunch of issues in this area and I really don't want to do anything that opens up a new (or old) security issue.

api/src/main/java/jakarta/servlet/ServletContext.java
markt-asf: Sorry - forgot to click the add button for this comment.

Arrgghhhh! Just when I think there is a workable solution in sight, I realise I have forgotten about one of the requirements. I think the suggestions above would give us a workable solution that removes most (hopefully all) the ambiguity in the current API. What it doesn't address is providing access to the path parameters and/or the 'raw' versions of contextPath, servletPath and pathInfo.

What if, rather than introducing the concept of modes, we added methods to `HttpServletMapping`. I'm thinking:
- `getContextPathEncoded()`
- `getServletPathEncoded()`
- `getPathInfoEncoded()`

where only the following had been performed;
- extract query string, if any, for separate processing
- decode unreserved %nn
- normalize
- reject dot-segemts with path parameters

gregw:
@markt-asf I'm going to do a meta response first before diving into the details of your reply.   I think we need to agree a general approach and then deep dive to stress test it.

I like your 'suspicious' sub-catagory of 'scary'.  The key sub-catagory in Jetty's handling is `ambiguous`, with the key example being decoding %2F where it is impossible for the app to tell if it is a real '/' without looking at the requestURI.   We also catagorize things like '%2e%2e' as ambiguous as currently the spec is a bit vague about what normalization should be done, specially for requests that may transit a RequestDispatcher (or 2) so an app can have a hard time knowing if a '..' is really a '..' or the result of decoding, double decoding or even tripple decoding etc.    

But regardless of the exact membership of the catagories, I'm seeing a general approaches we are discussing fitting into the following pattern/options:

 + existing path providing methods will have their exact decoding,normalization and de-parameterization defined.   For the existing methods they will:
     - never return undecoded characters in a path.  A % character in the path could only be there by being decoded from %25(including from RequestDispatcher).
     - never contain path parameters. a ; character must have been encoded (including from RequestDispatcher).
     - never contain a real '..' or '.' segment and so if they are present, they must have been encoded (including from RequestDispatcher)
 + default 6.0 behavior will be that path providing methods will not return 'ambiguous' paths.   This will be prevented either by:
    a) rejecting the request with a 400 before handling; or
    b) throwing an exception if the decoded path methods are called.
 + A container may also be configured to reject  'suspicious' paths.
 + Apps that want 'suspicious' or 'ambiguous' paths can:
    c) configure the container so that these methods may return decoded suspicious and/or ambigous paths (modal approach); and/or
    d) use new methods that are never ambiguous (method approach)
 + methods that take paths:
    e) will be default reject any paths that are not allowed to be returned by the current settings for the path providing methods (modal approach); and/or
    f) new method that take specifically encoded paths will be defined that avoid ambiguities.

I'm not saying we do all of a-f, but that our solution is probably a subset of a-f.

The approach a) feels better that b), except that it doesn't work well with new methods approaches d) and/or f).  If an app is safe because it is using the new methods then we should not reject with a 400 so they can use the new methods... but then if they have one component that calls the old methods they could be unsafe.   So perhaps the default rejection is a), but an app can declare it wants b) so that it can access new methods.

The advantage of c) modal approach is that we can match the behavior of the methods taking paths e) so that they only accept what can be produced.  
Conversely the problem of d) new methods is that the methods that take paths cannot know from where a passed string came.  It could be made from old methods, new methods, the app, the user or a brain-blowing combination of all! 

You know... I think we are perhaps giving ourselves an impossible task.    If we want an unambiguous safe well specified string representation of a URI, then that is a normalized encoded URI.  I don't think we can come up with a string representation any better than that.    Perhaps it would be better if any new methods were not string based.  We already have the (badly named) `getRequestURI` method that returns the URI string and can be passed to whatever URI decoding library the application wishes to use.

So rather than trying to invent a new way to make a safe string URI representation, we should just prevent the exising methods from working with unsafe string URI representations.    Then we provide new methods that are not based on strings, but rather classes.  We are after all programming in a strongly typed language.      Perhaps these methods could use the existing java.net.URI class, but I don't think that is very helpful, as it really just provides help with the structure of an absolute URI and for the path segment is basically just is a well checked string, but with all the same issues on `getPath()` vs `getRawPath()` and no smarts about HTTP paths (eg parameters).   What we need is a `HttpURIPath` class that provides safe and conveniant access to the path, segment by segment split into matched segments and info segments etc.    A new class like that would stop the brain explosions resulting from any method taking a string path, whose providence is entirely unknown.

In fact, the idea of a new class is kind of  just aggregatting the new methods suggested by @markt-asf into a class that can have a lot more convenience methods and use modern language features (Optional, Streams etc.) for navigating the segments/mapping of a URI.

 I'll stop at this point, as this is already a too long response...  but is the idea of a new class a non-starter or something I should flesh out a bit more?  

Sigh... just realized we need to include `HttpServletMapping.getMatchValue()` in our considerations!

markt-asf:
I think the idea of a new class is a good one.

I also agree with your general approach for the existing methods and much of the detail as well although there are a few places where we aren't quite aligned.

The existing methods are what I am most concerned about at this point. As much as it pains me to suggest it, given the time constraints around Servlet 6.0, how do you feel about a call to try and work through some of the detail? I'm thinking you, me and any other interested parties with the aim of bringing a proposal back here for wider feedback.

gregw:
@markt-asf I think a call would be good to progress this, as time is short.     But I think we need to do a bit of preparation for it, so all coming in are up to speed with the exact issues and we can focus on solution.  So can we make the call later in the week - Thursday or Friday?     Timezone might be a bit of an issue, although @stuartwdouglas  and I are in the same zone.  Is there anybody from a European timezone that wishes to attend?

What I'd like to do is move the preparatory discussion back to #18 itself rather than here on this PR, as I think no matter what there will be changes needed to this PR and we should not bias the discussion one way or another.    So today I will try to write a concise summary of the exact issue together with some example scary URIs.  I'll then try to put together a series of criteria/questions against which we can evaluate all the potential solutions.  I'll post something to #18 shortly.      Should I close this PR for now?

markt-asf:
I am in UTC+1 but can go late into my evening. Thursday works for me. No objections to moving the discussion back to #18. I think closing this PR for now makes sense.

PR #423, opened 2021-10-06
Add generic attribute support to SessionCookieConfig

markt-asf:
Also adds an entry to the change log.
The schema also requires an update. A separate PR will be provided shortly to the schema project for that change.


api/src/main/java/jakarta/servlet/SessionCookieConfig.java
```diff
@@ -231,4 +233,45 @@
...
+     * @param name Name of attribute to return
+     *
+     * @return Value of specified attribute
+     *
+     * @since Servlet 6.0
+     */
+    public String getAttribute(String name);
+
+    /**
+     * Obtain the Map (keys are case insensitive) of all attributes and values, including those set via the attribute
```
gregw, line 269: If the keys are case-insensitive (must they be?) then that needs to be mentioned in the javadoc of both `setAttribute` and `getAttribute`.

Rather than case-insensitive, could we just convert all the keys to lowercase?

markt-asf: Done.

PR #424, opened 2021-10-07
Issue #18 URI path processing

gregw:
Issue #18 URI path processing. Moved the wiki text to a PR for better comment and suggestions.

I've kept the TODOs in for now.  I will replace with PR comments and then remove.

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
 the `getDateHeader` method cannot translate the header to a `Date`
 object, an `IllegalArgumentException` is thrown.
 
+=== Request URI Processing
+The process described here adapts and extends the URI canonicalization process described in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) to create a standard Servlet URI path canonicalization process that ensures that URIs can be mapped to Servlets, Filters and security constraints in an unambiguous manner. It is also intended to provide information to reverse proxy implementations so they are aware of how requests they pass to servlet containers will be processed.
+
+Servlet containers may implement the standard Servlet URI path canonicalization in any manner they see fit as long as the end result is identical to the end result of the process described here. Servlet containers may provide container specific configuration options to vary the standard canonicalization process. Any such variations may have security implications and both Servlet container implementors and users are advised to be sure that they understand the implications of any such container specific canonicalization options.
+
+==== URI Transport
+===== HTTP/1.1
+The URI is extracted from the `request-target` as defined by [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.1). URIs in `origin-form` or `asterisk-form` are passed unchanged to stage 2. URIs in `absolute-form` have the protocol and authority removed to convert them to `origin-form` and are then passed to stage 2. URIs in `authority-form` are outside of the scope of this specification.
+
+===== HTTP/2
+The URI is the `:path` pseudo header as defined by [RFC 7540](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3) and is passed unchanged to stage 2.
+
+===== Other protocols
+Containers may support other protocols. Containers should extract an appropriate URI for the request from the protocol and pass it to stage 2.
+
+==== Separation of path and query
+The URI is split by the first occurrence of any '?' character to path and query.  The query is preserved for later handling and the following steps applied to the path.
+
+==== Discard fragment
+A fragment in the path is indicated by the first occurrence of a `\#` character.  Any `#` character and following fragment is removed from the path and discarded.
+
+==== Decoding of non-special characters
+Characters other than `/`, `;` and `%` that are encoded in `%nn` form are decoded and the resulting octet sequences is treated as UTF-8 and converted to a character sequence.
+
+> Note that special characters cannot be part of a UTF-8 character sequence as all such sequences are comprised of negative octets.
+
+> Note this is not reserved characters as defined by RFC3986, as that does not include `%` and includes many characters we don't care about. Avoiding a second decoding is worthwhile.
+
+==== Collapse sequences of multiple `"/"` characters
+> **WARNING** Swapping the order of stage 3 and stage 4 may be significant. Consider `"/aaa/bbb//../"`.
+
+> **TODO** Are we sure we don't want to do this in the other order?
+
+Any sequence of more than one `"/"` character in the URI must be replaced with a single `"/"`.
+
+==== Remove dot-segments+
+* A path not starting with "/" must be rejected with a 400 response.
+* Sequences of the form `"/./"` must be replaced with `"/"`.
+* Sequences of the form `"/" + segment + "/../"` must be replaced with `"/"`.
+* If there is no preceding segment for a `".."` segment then return a 400 response.
+
+==== Removal of path parameters
+A path segment containing the `";"` character is split at the first occurence of `";"`. The segment is replaced by the character sequence preceeding the `";"`. The characters following the `";"` are considered a path parameters and may be preserved by the container for later processing (eg `jsessionid`).
+
+> TODO How do we handle URIs like `/foo/;/bar`?  I think as currently written we end up with `/foo//bar` ?
+
+==== Decoding of remaining `%nn` sequences
+Any remaining `%nn` sequences in the path should be decoded. Some containers may be configured to leave some specific characters encoded (eg. the characters '/' and '%' may be left decoded by some container configuration).
+
+==== Mapping URI to context and resource
+The decoded path is used to map the request to a context and resource within the context. This form of the URI path is used for all subsequent mapping (web applications, servlet, filters and security constraints).
+
+==== Rejecting Suspicious Sequences
+If suspicious sequences are discovered during the prior steps, the request must be rejected with a 400 bad request using the error handling of the matched context.  By default the set of suspicious sequences includes:
+
+ * The encoded `"/"` character
+ * Any `"."` or `".."` segment that had a path parameter
+ * Any `"."` or `".."` segment with any encoded characters
+ * The `"\"` character encoded or not.
+ * Any control characters either encoded or not.
+
+A container or context may be configured to have a different set of rejected sequences.
+
+> TODO how do we define control characters? < 0x20? Is 0x7F (DEL) OK?
+
+> TODO should we also by default reject '\' and '%5c' ?
+
+> TODO should we also by default reject non visible and/or control characters ?
+
+> TODO if %2F is allowed, we may now have double '/', '/./' and '/../' segments in the URL, should stage 3 and 4 be re-run if this is allowed?
+
+
+
+
+
+
 === Request Path Elements
 
 The request path that leads to a servlet


spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+===== Other protocols
+Containers may support other protocols. Containers should extract an appropriate URI for the request from the protocol and pass it to stage 2.
+
+==== Separation of path and query
+The URI is split by the first occurrence of any '?' character to path and query.  The query is preserved for later handling and the following steps applied to the path.
+
+==== Discard fragment
+A fragment in the path is indicated by the first occurrence of a `\#` character.  Any `#` character and following fragment is removed from the path and discarded.
+
+==== Decoding of non-special characters
```
gregw, line 1324: Maybe this should be
```suggestion
==== Decoding of non-special octets as UTF-8
```
We can safely decode UTF-8 here as any visible US_ASCII character is never part of a UTF-8 multichar sequence.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+Characters other than `/`, `;` and `%` that are encoded in `%nn` form are decoded and the resulting octet sequences is treated as UTF-8 and converted to a character sequence.
+
+> Note that special characters cannot be part of a UTF-8 character sequence as all such sequences are comprised of negative octets.
+
+> Note this is not reserved characters as defined by RFC3986, as that does not include `%` and includes many characters we don't care about. Avoiding a second decoding is worthwhile.
+
+==== Collapse sequences of multiple `"/"` characters
```
gregw, line 1331: **WARNING** Swapping the order of stage 3 and stage 4 may be significant. Consider `"/aaa/bbb//../"`.

Are we sure we don't want to do this in the other order?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+The decoded path is used to map the request to a context and resource within the context. This form of the URI path is used for all subsequent mapping (web applications, servlet, filters and security constraints).
+
+==== Rejecting Suspicious Sequences
+If suspicious sequences are discovered during the prior steps, the request must be rejected with a 400 bad request using the error handling of the matched context.  By default the set of suspicious sequences includes:
+
+ * The encoded `"/"` character
+ * Any `"."` or `".."` segment that had a path parameter
+ * Any `"."` or `".."` segment with any encoded characters
+ * The `"\"` character encoded or not.
+ * Any control characters either encoded or not.
```
gregw, line 1362: Do we want this? If so, how do we define control characters?  < 0x20? Is 0x7F (DEL) OK?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+==== Decoding of remaining `%nn` sequences
+Any remaining `%nn` sequences in the path should be decoded. Some containers may be configured to leave some specific characters encoded (eg. the characters '/' and '%' may be left decoded by some container configuration).
+
+==== Mapping URI to context and resource
+The decoded path is used to map the request to a context and resource within the context. This form of the URI path is used for all subsequent mapping (web applications, servlet, filters and security constraints).
+
+==== Rejecting Suspicious Sequences
+If suspicious sequences are discovered during the prior steps, the request must be rejected with a 400 bad request using the error handling of the matched context.  By default the set of suspicious sequences includes:
+
+ * The encoded `"/"` character
```
gregw, line 1358: if %2F is allowed, we may now have double '/', '/./' and '/../' segments in the URL, should stage 3 and 4 be re-run if this is allowed?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,59 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
 object, an `IllegalArgumentException` is thrown.
 
+=== Request URI Processing
+The process described here adapts and extends the URI canonicalization process described in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) to create a standard Servlet URI path canonicalization process that ensures that URIs can be mapped to Servlets, Filters and security constraints in an unambiguous manner. It is also intended to provide information to reverse proxy implementations so they are aware of how requests they pass to servlet containers will be processed.
+
+Servlet containers may implement the standard Servlet URI path canonicalization in any manner they see fit as long as the end result is identical to the end result of the process described here. Servlet containers may provide container specific configuration options to vary the standard canonicalization process. Any such variations may have security implications and both Servlet container implementors and users are advised to be sure that they understand the implications of any such container specific canonicalization options.
+
+==== URI Transport
+===== HTTP/1.1
```
gregw, line 1309: We should probably also describe at least HTTP/1.0 and maybe HTTP/0.9

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: I don't like doing '/' collapsing before dot-segments, as it is a lot further from the RFC.
Perhaps we can just say that any `..` segment that consumes an empty segment is suspicious and thus rejected. Then we can collapse '//' after dot segment removal:

* `/foo//bar` would result in an allowed request to `/foo/bar`
* `/foo//../bar` would result in a rejected request to `/foo/bar`

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,59 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+
+==== Mapping URI to context and resource
+The decoded path is used to map the request to a context and resource within the context. This form of the URI path is used for all subsequent mapping (web applications, servlet, filters and security constraints).
+
+==== Rejecting Suspicious Sequences
+If suspicious sequences are discovered during the prior steps, the request must be rejected with a 400 bad request using the error handling of the matched context.  By default the set of suspicious sequences includes:
+
+ * The encoded `"/"` character
+ * Any `"."` or `".."` segment that had a path parameter
+ * Any `"."` or `".."` segment with any encoded characters
```
gregw, line 1350:
```suggestion
 * Any `"."` or `".."` segment with any encoded characters
 * Any `".."` segment preceded by an empty segment 
```

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+ * The `"\"` character encoded or not.
```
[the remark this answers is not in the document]
markt-asf: Not sure what you mean here. I think this needs to be in the list as it avoids a lot of complexity when we get to mapping URIs to file systems. Tomcat has been excluding it by default for ages without issue.

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: I'm not sure we need this. Their presence would be unusual but I don't see a good reason to exclude them by default at this point. (I'll note that if the URIs gets passed to the default servlet it might reject them with a 404 if there character is not a legal file name character for the underlying file system).

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,59 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+ * Any control characters either encoded or not.
+
+A container or context may be configured to have a different set of rejected sequences.
+
```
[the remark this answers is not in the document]
markt-asf: +1 - especially the various edge cases we've been discussing

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+Containers may support other protocols. Containers should extract an appropriate URI for the request from the protocol and pass it to stage 2.
+
+==== Separation of path and query
+The URI is split by the first occurrence of any '?' character to path and query.  The query is preserved for later handling and the following steps applied to the path.
+
+==== Discard fragment
+A fragment in the path is indicated by the first occurrence of a `\#` character.  Any `#` character and following fragment is removed from the path and discarded.
+
+==== Decoding of non-special characters
+Characters other than `/`, `;` and `%` that are encoded in `%nn` form are decoded and the resulting octet sequences is treated as UTF-8 and converted to a character sequence.
```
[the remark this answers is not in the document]
markt-asf: ';' is not the only character that can be used for path segments.

You cannot avoid two separate passes for %nn decoding if you want to be RFC 3986 compliant w.r.t path parameters and to able to support the use of sub-delims in resource paths.

I do accept that the original text (decode `unreserved`) doesn't cover UTF-8 sequences representing code points U+0080 and above.

I think the character we do care about are `sub-delims` (since they can all be used for path parameters), `%2F` (unless it has already been rejected which it will have been by default) and %25 (since % is the escape). 

How about:

>decode (as UTF-8) %nn sequences apart from those representing `sub-delims` and `%`.

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: I don't think '%2F' should ever be decoded - primarily because of the issues you describe above. I'm happy to allow enough leeway in the specification for containers to do this if they want but I don't see Tomcat doing any more than allowing `%2F` through in encoded form.

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: Depends on outcome of discussion on collapsing multiple `/`.

spec/src/main/asciidoc/servlet-spec-body.adoc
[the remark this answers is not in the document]
gregw: But we don't act on any of the other sub-delims, so there is no need to keep them encoded only to decode them afterwards.    We will have already processed the special meanings of `:`, `?` and '#` by the time we get to this step.  The only other characters we parse specially after this point are `;` and `%`, so I think they are the only ones we need to leave encoded (so if anything I think we can drop `\` from here).

If we don't decode chars like `&` or `=` here, then we will just decode them later and the results we be the same. 

Hmmm perhaps `=` might be a little bit special for the processing of path params.   Actually I have an idea for an alternate formulation of the algorithm that might be a lot clearer (but perhaps further away from how it might actually be implemented).  Standby ....

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,63 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+==== Remove dot-segments+
+* Sequences of the form `"/./"` must be replaced with `"/"`.
+* Sequences of the form `"/" + segment + "/../"` must be replaced with `"/"`.
+
+==== Removal of path parameters
+A path segment containing the `";"` character is split at the first occurence of `";"`. The segment is replaced by the character sequence preceeding the `";"`. The characters following the `";"` are considered a path parameters and may be preserved by the container for later processing (eg `jsessionid`).
+
+==== Decoding of remaining `%nn` sequences
+Any remaining `%nn` sequences in the path should be decoded. Some containers may be configured to leave some specific characters encoded (eg. the characters '/' and '%' may be left decoded by some container configuration).
+
```
gregw, line 1342: I think that rather than describing an algorithm that has the effect that we want regarding segments, we should be more direct and describe an algorithm in terms of segments.    If we ultimately make a new segment based API available, then it can be in terms of the segments described here and we don't need to re-specify.

```suggestion
==== Split path into segments
The path is split into multiple segments divided by the `/` character.

==== Remove path parameters
Any segment containing the `";"` character is split at the first occurrence of `";"`. The segment is replaced by the character sequence preceding the `";"`. The characters following the `";"` are considered a path parameters and may be preserved by the container for later decoding and/or processing (eg `jsessionid`).

==== Decode characters
Characters that are encoded in `%nn` form are decoded in each segment. The resulting octet sequences is treated as UTF-8 and converted to a character sequence.

==== Remove dot-segments
All segments that are exactly `"."` are removed from the segment series. Segments that are exactly '".."` and that are preceded by a non `".."` segment are removed together with the preceding segment.     This canonicalization differs from RFC3986 in that segments with `%` encoded dots and/or parameters may be treated as dot segments.

==== Remove Empty Segments
By default, empty segments are removed.  Containers may be configured to retain empty segments.

==== Concatenate segments
The segments are concatenated into a single path string with each segment preceded by the `"/"` character.  If a segment contains the `"/"` or `"%"` characters, and the request will not be rejected below, then the container may re-encode those characters to `%nn` form.
```

Param removal is before the decoding step, which is before the dot removal step.  This is the main deviation from the RFC, so I have noted it in the text.

Note that it is immaterial if an implementation actually decodes then re-encodes `%2F` or just leaves it encoded.  I think it reads better this way, as it makes leaving `%2F` in the exception case.

I have switched dot segment removal before empty segment removal. My reason is that if a container wishes to retain empty segments then the result is more consistent with RFC3986.

Note this algorithm "forgets" if a path did not start with '/'.  I don't think that is a problem as we can probably specify that it must start with a '/' in step 1, and then add a rejection below as defence in depth.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1300,6 +1300,82 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+Any sequence of more than one `"/"` character in the URI must be replaced with a single `"/"`.
+
+==== Remove dot-segments+
+* A path not starting with "/" must be rejected with a 400 response.
+* Sequences of the form `"/./"` must be replaced with `"/"`.
+* Sequences of the form `"/" + segment + "/../"` must be replaced with `"/"`.
+* If there is no preceding segment for a `".."` segment then return a 400 response.
+
+==== Removal of path parameters
+A path segment containing the `";"` character is split at the first occurence of `";"`. The segment is replaced by the character sequence preceeding the `";"`. The characters following the `";"` are considered a path parameters and may be preserved by the container for later processing (eg `jsessionid`).
```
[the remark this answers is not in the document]
gregw: See the proposed big rewrite below, which removes path params before dealing with empty segments.  I'll resolve this comment for now and add this example to the table of troublesome URIs

spec/src/main/asciidoc/servlet-spec-body.adoc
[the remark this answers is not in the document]
gregw: I've added the table, but it needs a lot of review... probably cut down a bit as well.

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: I'm kind of concerned about characters like `%00` that can fool any native code handling.
Also  `%08`, `%0D` and `%7F` can be used to hide characters in a displayed URI (backspace, carriage ret and delete).

Perhaps if we don't reject these by default, we use this as an example of extra rejections that a container might do?

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: OK. I've switched back the order in my proposal below to have empty segment removal before dot dot resolution.

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: @markt-asf do you think there is enough leeway in my alternative text that is more segment based?  I know it says decode and then re-encode, but that is really for clarity and an implementation would be free to leave encoded.   The text says that if the %2f is allowed by the container then it "may" re-encode (or leave encoded).  Happy to make that a "should".

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: not needed

spec/src/main/asciidoc/servlet-spec-body.adoc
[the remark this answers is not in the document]
gregw: New proposal below.  I'll resolve this discussion now for clarity and we can discuss on the other proposal

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: I'm going to merge this suggestion to the PR now, because I think we have resolved most of the behaviour differences and it is now just a matter of how we describe it in text.  I'd like to do some cleanup of this and it is getting a bit meta editing a suggestion of a PR....   
Merging doesn't mean I think this text is settled and still happy to discuss and/or revert to previous form if that is the consensus.

stuartwdouglas:
Looks like the PR is a bit messed up, it has the head content length stuff in it now.

spec/src/main/asciidoc/servlet-spec-body.adoc
stuartwdouglas: I am ok with rejecting by default. I think the actual use cases for this are pretty much non-existent, so if they are present there is almost certainly something nefarious going on.

gregw:
@stuartwdouglas DOH! I did a merge from master and my git-foo must have messed up.   Let me see if I can fix....

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: @stuartwdouglas The third option when `/%2e%2e/` is allowed is to throw if `getPath()` or `getPathInfo()` are called.   That way it is only allowed if `getRequestUri()` or new API is used.  Actually `getRequestUri()` can still be dangerous with this... so perhaps if it is allowed it is only for new API?

If that is the case, we can move decoding until after dot-segment handling.

Note as this comment is outdated. I'm going to resolve and then re-comment on latest text.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,139 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+| /secret%2Fprivate.xml | /secret%2Fprivate.xml | 400
+| /secret%5Cprivate.xml | /secret\private.xml | 400
+| /secret%00/private.xml | /secret␀/private.xml | 400? 
+| /./secret/private.xml | /secret/private.xml| 403
+| /.;/secret/private.xml | /secret/private.xml| 403
+| /%2e/secret/private.xml | /secret/private.xml | 400
+| /%2e;/secret/private.xml | /secret/private.xml  | 400
+| /public/../secret/private.xml | /secret/private.xml | 403
+| /public/..;/secret/private.xml | /secret/private.xml | 403
+| /public/%2e%2e/secret/private.xml | /secret/private.xml | 400
```
gregw, line 1429: @stuartwdouglas said:
> So turns out I actually hate both the options when it comes to /%2e%2e/ being allowed :-)
> As written where we have different behavior to the RFC if you have a path like:
> `/public/%2e%2e/private/secret.txt`
> The end result will be:
> `/private/secret.txt` (which is different to what we had discussed before, where we were going to map to the context first). This is a security issue if a front end proxy is securing /private
> If we do it the other way around and we end up with:
> `/public/../private/secret.txt`
> This will map to /public as expected, but you end up with a path that starts with a .. segment, which these rules are largely supposed to prevent.
> I guess this is ok because %2e%2e is disallowed by default, but things seem to get very problematic when it is allowed.

gregw: I tend to agree.
I think that if `/%2e%2e/` segments are allowed then, by default, we should throw if and of the existing path methods are used and thus they will only be available from `getRequestUri()` or new API.

If we did this, then we could decode after dot-segment removal and be much closer to the RFC.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,139 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+  * Any path starting with an initial segment of `".."`
+  * The encoded `"/"` character
+  * Any `"."` or `".."` segment that had a path parameter
+  * Any `"."` or `".."` segment with any encoded characters
+  * Any `".."` segment preceded by an empty segment
+  * The `"\"` character encoded or not.
+  * Any control characters either encoded or not.
+
+==== Example URIs
+> TODO more work needed on this section. Perhaps some parts need to be moved to other sections.
```
gregw, line 1378: Once we are a bit more agreed on the algorithm, I think we should write it in java in `src/test/java` and the run all the examples below through it and generate this table.

gregw: I've pushed first attempt at this. The table is now generated from the algorithm

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
++
+The URI is split by the first occurrence of any '"?"' character to path and query.  The query is preserved for later handling and the following steps applied to the path.
+
+. **Discard fragment.**
++
+The path is split by the first occurrence of any `"\#"` character. The `"#"` and following fragment are discarded and the path is replaced with the character sequence preceding the `"#"` character.
+
+. **Split path into segments.**
++
+The path is split into multiple segments divided by the `"/"` character. A leading `"/"` does not indicate an initial empty segment.
```
markt-asf, line 1338: I don't understand what you mean by

> A leading `"/"` does not indicate an initial empty segment.

Ah. Do you mean the `/`s used to split the path into segments are not included in the resulting segments?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+
+Servlet containers may implement the standard Servlet URI path canonicalization in any manner they see fit as long as the end result is identical to the end result of the process described here. Servlet containers may provide container specific configuration options to vary the standard canonicalization process. Any such variations may have security implications and both Servlet container implementors and users are advised to be sure that they understand the implications of any such container specific canonicalization options.
+
+. **Separation of path and query.**
++
+The URI is split by the first occurrence of any '"?"' character to path and query.  The query is preserved for later handling and the following steps applied to the path.
+
+. **Discard fragment.**
++
+The path is split by the first occurrence of any `"\#"` character. The `"#"` and following fragment are discarded and the path is replaced with the character sequence preceding the `"#"` character.
```
markt-asf, line 1334: The fragment occurs *after* the query string so will already have been removed in "Separation of path and query" stage. "Discard fragment" should be moved to before "Separation of path and query".

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+
+. **Split path into segments.**
++
+The path is split into multiple segments divided by the `"/"` character. A leading `"/"` does not indicate an initial empty segment.
+
+. **Remove path parameters.**
++
+Any segment containing the `";"` character is split at the first occurrence of `";"`. The segment is replaced by the character sequence preceding the `";"`. The characters following the `";"` are considered path parameters and may be preserved by the container for later decoding and/or processing (eg `jsessionid`).
```
markt-asf, line 1342: RFC 3986 allows any sub-delim to be used to delimit path parameters. `;` is certainly the most popular. Do we want to limit the servlet spec to just `;`? The performance/complexity cost of replacing `;` with `sub-delim` in this algorithm looks to be pretty minimal and the benefit is we avoid a deviation from RFC 3986.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
++
+Characters that are encoded in `%nn` form are decoded in each segment. The resulting octet sequences is treated as UTF-8 and converted to a character sequence that replaces the segment.
+
+. **Remove Empty Segments.**
++
+Empty segments are removed.  Containers may be configured to retain empty segments.
+
+. **Remove dot-segments.**
++
+All segments that are exactly `"."` are removed from the segment series. Segments that are exactly `".."` and that are preceded by a non `".."` segment are removed together with the preceding segment.     This normalization differs from RFC3986 in that segments with `%` encoded dots and/or parameters may be treated as dot segments.
```
markt-asf, line 1354: I've seen this view (`/%2e%2e/` != `/../`) expressed in multiple places but I can find text in RFC 3986 to support it. Section 2.3 seems to say exactly the opposite. What am I missing?

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+
+. **Concatenate segments.**
++
+The segments are concatenated into a single path string with each segment preceded by the `"/"` character. If the original path after stripping query and fragment ends with a `"/"` then append a final `"/"` character.
+ If a segment contains the `"/"` or `"%"` characters, and the request will not be rejected below, then the container may configured re-encode those characters to the `%nn` form. If any characters are re-encoded, then the `"%"` must also be re-encoded.
```
markt-asf, line 1359: "may configured" - the wording isn't quite right.

"should" rather than "may"? It leaves the option not to re-encode open while strongly suggesting the safe approach.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+  * Any control characters either encoded or not.
+
+==== Example URIs
+> TODO more work needed on this section. Perhaps some parts need to be moved to other sections.
+
+Consider the following URIs with a forbidden security constraint at `/secret/*` and a servlet registered at /dispatch/* that forwards request to the string return from pathInfo()
+
+. Example URIs
+|===
+| Encoded URI path  | Decoded Path | Rejected
```
markt-asf, line 1385: Nearly all of these examples have `/` appended to the decoded path. That looks wrong.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,145 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+| `/public%2Ffile.txt` | `/public/file.txt/` | 400 encoded /
+| `/public%5Cfile.txt` | `/public\file.txt/` | 400 backslash character
+| `/public%00/file.txt` | `/public /file.txt/` | 400 control character
+| `/public/./file.txt` | `/public/file.txt/` |
+| `/public/%2e/file.txt` | `/public/file.txt/` | 400 encoded dot segment
+| `/public/%2e;/file.txt` | `/public/file.txt/` | 400 encoded dot segment
+| `/../docroot/public/file.txt` | `/../docroot/public/file.txt/` | 400 leading dot-dot-segment
+| `/public/dir/../file.txt` | `/public/file.txt/` |
+| `/public//../file.txt` | `/file.txt/` |
+| `/public/;param/../file.txt` | `/file.txt/` |
```
markt-asf, line 1401: Looking at this example made me uneasy. Can we add `"/" sub-delim *(pchar)` to the list of suspicious sequences to be rejected?

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: I don't think there is any reason for us to strip out any other sub-delims:
```      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="
```

If we suddenly start stripping and discarding the path after any of these characters, then I think we will break a lot of applications.  I think the RFC makes it clear that interpretation of these sub-delims is implementation and/or schema specific:

> URI producing applications
   often use the reserved characters allowed in a segment to delimit
   scheme-specific or dereference-handler-specific subcomponents.  For
   example, the semicolon (";") and equals ("=") reserved characters are
   often used to delimit parameters and parameter values applicable to
   that segment.  The comma (",") reserved character is often used for
   similar purposes.  For example, one URI producer might use a segment
   such as "name;v=1.1" to indicate a reference to version 1.1 of
   "name", whereas another might use a segment such as "name,1.1" to
   indicate the same.  Parameter types may be defined by scheme-specific
   semantics, but in most cases the syntax of a parameter is specific to
   the implementation of the URI's dereferencing algorithm.

We only act on ";", so I think we only need to treat ";" specially.

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: yep.   How about:
```suggestion
The path is split into multiple segments divided by the `"/"` character, ignore any leading `"/"`.
```

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: Ah very VERY good point.  I think I've tricked myself into thinking that because the RFC does allow reserved chars like `/` to be encoded to exclude them from normalization.... so I think I had assumed that `.` could also be encoded to exclude it from normalization.   But reading it again now, you're right, that `.` encoding must be decoded before normilization.       So we are mostly doing the right thing after all and I need not fret so much about it.      We still have a few differences, but not huge ones. I'll update text.

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: I already have 

> Any `".."` segment preceded by an empty segment

But that doesn't fire because we are removing empty segments before we normalize.   As said above, I'd prefer to remove empty segments last so this one would not be an issue.
But for now I'll add 

> Any empty segment with a parameter

Whilst we discuss the ordering.   I don't mind if ultimately we leave in suspect patterns that cannot happen by default, as they may still trigger if there are container specific configurations.  Thus containers will need to both configure normalization and explicitly allow suspect patterns for them to arrive at the application.

gregw: This is now:
> | `/public/;param/../file.txt` | `/file.txt` | 400 empty segment with parameters

But I would prefer
> | `/public/;param/../file.txt` | `/public/file.txt` | 400 empty segment before dot dot

gregw: Actually, I have convinced myself otherwise.  The reason we remove empty segments is that many file systems treat them the same as a '.' segment, therefore we should do the same.
So `/foo//bar` should be handled the same as `/foo/./bar` and be normalized to `/foo/bar`.    But normalization of `/foo/./../bar` by the RFC is (surprisingly) `/bar`, so `/foo//..bar` should normalize to the same result, as should `/foo/;/../bar`.

gregw:
I've done a moderate refactor of the code and examples.   I found a few edge cases that I don't think were handled correctly by the algorithm as specified/implemented (e.g. `/.` ended up as the empty string rather than `/`).      So I added a lot more tests for those.

markt-asf:
+1 for a call early this coming week. I should be available from 08.00 UTC any day or can go as late as 00.00 UTC if that is easier.

gregw:
@markt-asf @stuartwdouglas meeting invite sent.... I hope it is not too late for you @stuartwdouglas ?  I can probably do a couple of hours earlier if need be?

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: Ack. Agree that the Servlet spec will only allow ';' to delimit path parameters.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,169 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+| `/./foo/bar` | `/foo/bar` |
+| `/foo/%2e/bar` | `/foo/bar` | 400 encoded dot segment
+| `/foo/.;/bar` | `/foo/bar` | 400 dot segment with parameter
+| `/foo/%2e;/bar` | `/foo/bar` | 400 encoded dot segment
+| `/foo/.%2Fbar` | `/foo/./bar` | 400 encoded /
+| `/foo/.%5Cbar` | `/foo/.\bar` | 400 backslash character
+| `/foo/bar/.` | `/foo/bar` |
+| `/foo/bar/./` | `/foo/bar/` |
+| `/foo/bar/.;` | `/foo/bar` | 400 dot segment with parameter
+| `/foo/bar/./;` | `/foo/bar` | 400 empty segment with parameters
```
markt-asf, line 1411: This (and a couple of variations) are the only query I have at the moment. I think the decoded path should end with `"/"`.
(see above)

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,169 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
++
+Empty segments are removed.  Containers may be configured to retain empty segments.
+
+. **Remove dot-segments.**
++
+All segments that are exactly `"."` are removed from the segment series. Segments that are exactly `".."` and that are preceded by a non `".."` segment are removed together with the preceding segment. This normalization differs from RFC3986 in that segments with parameters may be treated as dot segments.
+
+. **Concatenate segments.**
++
+The segments are concatenated into a single path string with each segment preceded by the `"/"` character. If the original path after stripping query and fragment ends with a `"/"` or all segments have been removed, then append a final `"/"` character.
```
markt-asf, line 1356: I'd like to consider changing this to 

> ...after stripping any query, fragment and/or path parameters ends with a `"/"`...

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: +1
As should the ``/foo;/bar;/;`` above, which should also not be reject with with empty segment.   
I'm going to try to fix with last empty segment handling.... stand by....

gregw: I have pushed an alternative that treats any trailing `/` as an empty segment.  I think it is clearer and simpler?     Unit test now passes all the asserts

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: text no longer needed with empty segment changes.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,169 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+| `/foo/bar/..` | `/foo` |
+| `/foo/bar/../` | `/foo/` |
+| `/foo/bar/..;` | `/foo` | 400 dot segment with parameter
+| `/foo/bar/../;` | `/foo/` |
+| `/foo/..bar` | `/foo/..bar` |
+| `/foo/.../bar` | `/foo/.../bar` |
+| `/foo//bar` | `/foo/bar` |
+| `//foo//bar//` | `/foo/bar/` |
+| `/;/foo;/;/bar/;/;` | `/foo/bar/` | 400 empty segment with parameters
+| `/foo//../bar` | `/bar` |
```
lachlan-roberts, line 1432: This has been described above as one of the suspicious sequences which should be rejected with 400 response.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -1303,6 +1303,169 @@ the header value to an `int`, a `NumberFormatException` is thrown. If
...
+
+. **Decode.**
++
+Octets that are encoded in `%nn` form are decoded in each segment. The resulting octet sequence is treated as UTF-8 and converted to a character sequence that replaces the segment.
+
+. **Remove Empty Segments.**
++
+Empty segments, other than the last segment, are removed.  Containers may be configured to retain all empty segments.
+
+. **Remove dot-segments.**
```
lachlan-roberts, line 1350: Shouldn't the steps `Remove Empty Segments` and `Remove dot-segments` be swapped in order?

If the suspicious sequence check was disabled due to a different container configuration, then I would have expected that `/path//../info` should decode into `/path/info` not to `/info`.

gregw: I thought that at first also.   But ultimately empty segments are treated by the filesystem the same as `/./`.    The normalization of `/foo/./../bar` is `/bar/`, so the normalization of  `/foo//../bar` should also be `/bar/`.

spec/src/main/asciidoc/servlet-spec-body.adoc
gregw: yeah... but that was really only if containers don't remove empty segments.   I think that either needs to be made clear or that rejection removed.

gregw:
@markt-asf thanks for those updates.... they mostly look good.

However, the text reads as:

 > If a segment contains the `"/"` or `"%"` characters, and **the container is configured to not reject the request**, then the container should re-encode those characters to the `%nn` form.

which I think is correct, but that's not what we have currently implemented in the unit test.   We are always re-encoding these characters, so a URI of `/foo%25/bar` will always stay encoded and never be returned as `/foo%/bar`.    I'm not sure that is what we intended?

Perhaps we should change the text to:

 > If a segment contains the `"/"` or `"%"` characters, and **the container is configured to not reject the request for containing a %2F**, then the container should re-encode those characters to the `%nn` form.

And then we can make the test encoding conditional on rejecting the %2F and add an example for `/foo%25/bar`

gregw:
I ran this algorithm past our unit tests.  A few differences but only in things that I don't mind changing.
I think this PR is pretty close!

markt-asf:
I agree. I performed a similar test with Tomcat with similar results. The only thing I could see that was required was adding the canonicalization text to the other HttpServletReqiuest methods. I've just done that. I think this is good to merge.

gregw:
I'll wait for @stuartwdouglas to comment before merging.

Issue #427, opened 2021-10-19
Use jdk11 for servlet-api


gregw:
The API currently is both source and target 1.8.  This should be updated to at least 11.

PR #428, opened 2021-10-19
Update to jdk11

gregw:
For #427 update to jdk11


arjantijms:
Updated the stage scripts on the CDI to use JDK 11 (openjdk).

PR #434, opened 2021-11-03
Clarify under what circumstances onError is called

### LINKED ISSUE #433: Differences in calling WriteListener.onError
If an async servlet is in a loop like:
```java
    ServletOutputStream out = response.getOutputStream();
    out.setWriteListener(new WriteListener()
    {
        public void onWritePossible()
        {
            try
            {
                while (out.isReady())
                    out.write(someBuffer);
             }
             catch(Throwable t)
             {
                 t.printStackTrace();
             }     
        }
  
        public void onError(Throwable t)
        {
            t.printStackTrace();
        }
    });
```

and the client receiving the response closes the connection at some point, then we see different behaviors from Jetty, Tomcat and Undertow:

- Jetty throws from write and calls onError with a different exception.  If `isReady()` is called it returns true.
- Tomcat throws from write and calls onError with the cause of the original exception.  If `isReady()` is called it returns false.
- Undertow throws from the write, but never calls onError.  If `isReady()` is called it return false, but then `onError` is still not called.

I'm not sure any of these behaviors is strictly correct:
 + should onError be called for an exception that has already been thrown to the application?
 + should onError every be called if `isReady()` has not been called and has not returned false?
 + should the write every actually throw? instead should `isReady()` return false and then `onError` is called with the problem?
 + if the write has thrown, what should `isReady()` return?

stuartwdouglas:
Fixes #433


stuartwdouglas:
I have ended up making some changes compared to what was discussed, as the more I thought about it I thought there were some issues with the proposals we talked about on the issue. In particular I think the issues are:

- At the moment AsyncListener.onError is only really called as a result of dispatch operations, having it occur due to underlying connection failure (potentially delivered by an IO thread) seems like a breaking change, as this method could now be called in circumstances where it was not possible before.
- ReadListener and WriteListener onError methods are now invoked when a read is attempted (generally when isReady is called). This matches the behavior of blocking mode, where you only get notified of IO failure when you attempt to perform an operation.

Basically the more I thought about it the more I came to the conclusion that if we want to support eager notification of IO failure (e.g. RST_STREAM) we should add a special listener just for this that is usable in both blocking and non blocking mode. 

I also think that delivering `IOException` to AsyncListener on eager IO failure is also a mistake, as this is not what the method has been used for in the past.

gregw:
Stuart, 
I'll review your text in detail shortly, but I do have an issue with:

> * At the moment AsyncListener.onError is only really called as a result of dispatch operations, having it occur due to underlying connection failure (potentially delivered by an IO thread) seems like a breaking change, as this method could now be called in circumstances where it was not possible before.

It is a not uncommon use case that some compute intensive job is done asynchronous, but it is desirable to stop that work if the connection is ever closed.  This is an oft requested feature, but in the days of http1 was just not possible because you could not detect close without attempting IO.   But this is now possible with http2 and http3.  Jetty definitely has users that plug into AL.onError and extect to see some exception or other if the stream is reset, even if no IO is attempted.

Now that we have h2/h3 it seams silly to make such applications waste all that CPU finishing their compute intensive task, only to finally discover the connection has been closed when they try to write the response.   We would have to put in a non-compliant mode to support such users if we end up hiding exceptions without IO.      Now they don't need to be delivered to AL.onError... so long as they are delivered to one of the onError listeners in a timely fashion, then we are good.

stuartwdouglas:
The issue I have with this approach is that you can still have that requirement even with blocking IO and without using `startAsync` at all. If we tie this capability into the read/write listeners (or even the AL) then we limit the applications that can make use of it.

gregw:
Hmmm the text is rather complex and on my third read through I'm still not sure I entirely understand.

I'm wondering if we are trying to be too smart by avoiding multiple reports of the same issue.    What about if whenever there is an issue with the connection, we report it once to each and any of the onError listeners registered.
I can't really think of any exceptions that are going to affect the read side, but not the write side or versa vice.   If a read suffers a broken pipe exception, then we are going to have to onError the write side as well. If a h2 stream is reset, that is both read and write.

So how about:
 + read/write/close/flush do not throw, unless called after any onError callback has been called.
 + if an error happens on a connection or during any IO operation, then it will be reported to any WL, RL or AL registered (in that order!) and isReady will return false from that time forward.

stuartwdouglas:
Basically something like this: https://github.com/eclipse-ee4j/servlet-api/pull/435

stuartwdouglas:
> 
> So how about:
> 
>     * read/write/close/flush do not throw, unless called after any onError callback has been called.

+1, that should already be covered by the text

> 
>     * if an error happens on a connection or during any IO operation, then it will be reported to any WL, RL or AL registered (in that order!) and isReady will return false from that time forward.

I would rather deliver do both a WL and RL, as long as the relevant streams are still open (e.g. if read() has returned -1 then we should not deliver to the RL, same if close has been called on either stream).

If we can't deliver to a RL or WL then I would rather add a mechanism like in #435 that can always be notified even when async is not in use.

gregw:
@stuartwdouglas I'm OK with not calling WL.onError or RL.onError if they have been closed, which means:

- a RL is closed if:
  - InputStream.close() has been called
  - a read has returned -1 or 
  - onAllDataRead has been called
  - RL.onError has previously been called
- a WL is closed if:
  - OutputStream.close has been called
  - exactly the content length has been written (successfully)
  - WL.onError has previously been called

As for AL.onError, I'm happy for it to always be called, or if we want to avoid duplication, it could be called IFF one of the RL or WL is either not set or is closed.

stuartwdouglas:
I agree with your description of 'RL is closed', however 'WL is closed if OutputStream.close has been called' should be 'OutputStream.close has been called and any buffered data has been successfully written to the client', as the close could trigger a write which could then fail. 

Like I mentioned above I would much rather we add a connection specific listener rather than repurpose AL.onError to handle async connection failures. If everyone else is in favor I guess I can live with it but it feels wrong to me, and can change the behaviour of existing applications as it can now be called at times when it would not otherwise have been called.

markt-asf:
The current Tomcat behaviour during async IO can be summarised as:
- any async IO errors are delivered to WL (if configured) and RL (if configured)
- If RL.isReady() triggers an IOE (this is where Tomcat does the read) it is delivered RL.onError
- If RL.onDataAvailable throws any throwable it is delivered RL.onError
- If RL.onAllDataRead throws any throwable it is delivered RL.onError
- If SOS.close or SOS.flush throws an IOE it it delivered to WL.onError
- If WL.onWritePossible throws any throwable it is delivered WL.onError
- Any throwables from the app are delivered to AL.onError

On reflection, I share Stuart's concern about starting to send async connection issues to AL. I can see the benefits of an new listener for these.

I wonder if we can simplify the "is RL/WL closed" element to "if the request is in async mode"?

I think we should include the handling of exceptions thrown from RL.onDataAvailable and friends in the list of reasons RL/WL onError will be called.

gregw:
@stuartwdouglas Can you wake up this PR and let's get it merged.

api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -273,6 +273,15 @@ public void println(double d) throws IOException {
...
      * This method can be used to determine if data can be written without blocking.
+     * <p>
+     * If an attempt is made to write to the stream when the stream is in async mode and this method has not returned
+     * {@code true} the method will throw an {@link IllegalStateException}.
+     * <p>
+     * If an error occurs and {@link WriteListener#onError(Throwable)} is invoked then this method will always return false,
+     * as no further IO operations are allowed after {@code onError} notification.
+     * <p>
+     * Note that due to the requirement for {@code write} to never throw in async mode, this method must return false if a
+     * call to {@code write} would result in an exception.
```
gregw, line 284:
```suggestion
    * If an error occurs then either: this method will return {@code true} and 
    * an {@link IOException} will be thrown from a subsequent call to {@code write(...)};  
    * or this method will return {@code false} and subsequently {@link WriteListener#onError(Throwable)} 
    * will be invoked with the error.  Once the error has either been thrown or passed to 
    * {@link WriteListener#onError(Throwable)}, then this method will always return {@code true} and all 
    * further {@code write} operations will throw an {@link IOException}.
```

api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -283,8 +292,17 @@ public void println(double d) throws IOException {
...
+     * <p>
+     * Note that after this method has been called methods on this stream that are documented to throw {@link IOException}
+     * will no longer throw these exceptions directly, instead any exception that occurs will be reported via
+     * {@link WriteListener#onError(Throwable)}. Please refer to this method for more information. This only applies to
+     * {@code IOException}, other exception types may still be thrown (e.g. methods can throw {@link IllegalStateException}
+     * if {@link #isReady()} has not returned true).
+     * <p>
+     * Once this method has been called {@link #flush()} and {@link #close()} become asynchronous operations, they will be
+     * performed in the background and any problems will be reported through the {@link WriteListener#onError(Throwable)}
+     * method.
```
gregw, line 305:
```suggestion
     * Note that after this method has been called, methods on this stream that are documented 
     * to throw {@link IOException} might not throw these exceptions directly, 
     * instead they may be reported via {@link ReadListener#onError(Throwable)} after a call to 
     * {@link #isReady()} has returned {@code false}. 
     * Please refer to {@link #isReady()} method for more information. 
     * <p>
     * Once this method has been called {@link #flush()} and {@link #close()} become asynchronous 
     * operations, possibly performed in the background. Thus any problems may be reported through the 
     * {@link WriteListener#onError(Throwable)} method, which may be called at any time 
     * after a {@link #close()} or otherwise only after a call to {@link #isReady()} has returned
     * {@code false}.
```

api/src/main/java/jakarta/servlet/WriteListener.java
```diff
@@ -39,7 +39,19 @@ public interface WriteListener extends EventListener {
...
+     *
+     * <ul>
+     * <li>{@link ServletOutputStream#isReady()} has been invoked and returned false</li>
+     * <li>{@link ServletOutputStream#close()} has been called, and the failure occurred before the response could be fully
+     * written to the client</li>
+     * </ul>
+     *
+     * If these conditions are not met and the stream is still open then any failure notification will not be delivered
+     * until {@link ServletOutputStream#isReady()} is invoked. {@code isReady} must return false in this situation, and then
+     * the failure will be delivered to the {@link #onError(Throwable)} method.
```
gregw, line 54:
```suggestion
     * Invoked when an error occurs writing data after 
     * {@link ServletOutputStream#setWriteListener(WriteListener)} has been called. 
     * This method will be invoked if there is a problem while data is being written to the 
     * stream and either:
     * <ul>
     * <li>{@link ServletOutputStream#isReady()} has been invoked and returned false</li>
     * <li>{@link ServletOutputStream#close()} has been called, and the failure occurred before 
     * the response could be fully written to the client</li>
     * </ul>
     *
     * If these conditions are not met and the stream is still open then any failure
     * notification will be delivered either:
     * by an exception thrown from a {@code IO} operation after an invocation of 
     * {@link ServletOutputStream#isReady()} has returned {@code true}; or by a call to this method 
     * after an invocation of {@link ServletOutputStream#isReady()} has returned {@code false};
     * <p>
     * This method will not be invoked in any circumstances after 
     * {@link AsyncListener#onComplete(AsyncEvent)} has been called.
```

api/src/main/java/jakarta/servlet/ReadListener.java
```diff
@@ -48,7 +48,19 @@ public interface ReadListener extends EventListener {
...
+     *
+     * <ul>
+     * <li>{@link ServletInputStream#isReady()} has been invoked and returned false</li>
+     * <li>{@link ServletInputStream#close()} has not been called</li>
+     * <li>{@link ServletInputStream#read()} (or any other read method) has not returned {@code -1}</li>
+     * </ul>
+     *
+     * If these conditions are not met and the stream is still open then any failure notification will not be delivered
+     * until {@link ServletInputStream#isReady()} is invoked. {@code isReady} must return false in this situation, and then
+     * the failure will be delivered to this method.
```
gregw, line 63:
```suggestion
      * Invoked when an error occurs reading data after {@link #setReadListener(ReadListener)} 
      * has been called.  This method will be invoked if there is a problem while data is being 
      * read from the stream and either:
      * <ul>
      * <li>{@link ServletInputStream#isReady()} has been invoked and returned false;</li>
      * <li>or {@link ServletInputStream#close()} has been called, and the failure occurred 
      *       before the response could be completed</li>
      * </ul>
      * If these conditions are not met and the stream is still open then any failure
      * notification will be delivered either:
      * by an exception thrown from a {@code IO} operation after an invocation of 
      * {@link ServletInputStream#isReady()} has returned {@code true}; or by a call to this method 
      * after an invocation of {@link ServletInputStream#isReady()} has returned {@code false};
      * <p>
      * This method will not be invoked in any circumstances after 
      * {@link AsyncListener#onComplete(AsyncEvent)} has been called.
```

gregw (changes requested):
updates in line with my comments on the issue #433

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -8546,6 +8546,9 @@ Jakarta Servlet {spec-version} specification developed under the Jakarta EE Work
 
 === Changes Since Jakarta Servlet 5.0
 
+link:https://github.com/eclipse-ee4j/servlet-api/issues/433[Issue 433]::
+Clarify how IO errors are handled by async streams.
+
```
markt-asf, line 8551: This needs moving to the correct section

markt-asf:
I think I am OK with the direction this is heading in but I'd like to review the changes once the PR has been rebased to take account of the clarifications that have already been made to the affected classes.

stuartwdouglas:
Sorry, I was on PTO, I have rebased and applied Greg's suggestions.

spec/src/main/asciidoc/servlet-spec-body.adoc
stuartwdouglas: Moved

api/src/main/java/jakarta/servlet/WriteListener.java
```diff
@@ -40,7 +40,21 @@ public interface WriteListener extends EventListener {
     void onWritePossible() throws IOException;
 
     /**
-     * Invoked when an error occurs writing data using the non-blocking APIs.
+     * Invoked when an error occurs writing data after {@link ServletOutputStream#setWriteListener(WriteListener)} has been
+     * called. This method will be invoked if there is a problem while data is being written to the stream and either:
+     * <ul>
+     * <li>{@link ServletOutputStream#isReady()} has been invoked and returned false</li>
```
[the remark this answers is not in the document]
stuartwdouglas: I also really don't like the idea of doubling up error handling. The original proposal meant that you only had to implement `onError`, with this change you now need to handle errors thrown from the stream. Should we add a section that if `onWritePossible` throws `IOException` then the `onError` method is called? It seems like obvious behavior but I am not sure if it is called out anywhere. This would mean that if `write` throws you can just let the exception propagate and the listener will handle it.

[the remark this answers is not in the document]
gregw: Stuart, by `onError` in your comment, I'm assuming that you mean `AsyncListener.onError` and not `WriteListener.onError`.

So I do like the "if `OWP` throw then the `AL.onError` method is called" as a good way to give control to the application about how write errors are reported to `AL.onError`.

I agree it there is something strange about not reporting a known error to `WL.onError` until `isReady()` is called.  But if we do not, then we do not have an easy way to tell if a previous write has completed or not.  Currently the only way we have on knowing a previous write has completed is if `isReady()` has returned true.   Perhaps you could also say that if `onWritePossible` has been called, that also indicates completion... but really `isReady()` should be called from within OWP and checked for a true response to protect from spurious wakeups.   If `WL.onError` can be called at any time, then it may be called simultaneously to another thread calling `isReady()` and then the app will never know if the call to `WL.onError` was the result of the false return from `isReady` or if it just happened anyway and another call is on its way.

I'm not sure there is a good solution given the current API.  I think the best we can do is be rigorous with the scheduling so we at least avoid races like the one above.

Note that if the app you described really wants  to know about an error before the next  write is ready, there is nothing stopping it calling `isReady()` immediately after the `write` and then it will know that the operation has either completed or that either ODA or WL.onError will be called as soon as the operation is complete.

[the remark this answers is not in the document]
stuartwdouglas: Thinking about it some more I agree with your concerns about thread safety. The only way I can think to make this work is to allow this use case via the flush method. We could add something like 'If flush() is called in async mode then isReady must not return true until the data is written out to the client'. 

Then if you really care about error notification and are not going to immediately write again you can call `flush + isReady` to see the results.

PR #437, opened 2021-11-05
Fixes generation of javadoc using JDK 11

arjantijms:

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -291,13 +291,7 @@
                         </goals>
                         <configuration>
                             <quiet>true</quiet>
-                            <source>8</source>
-                            <!-- 
-                                Needed to prevent the following error:
-                                The code being documented uses packages in the unnamed module, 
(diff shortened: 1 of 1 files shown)


markt-asf:
Which JDK are you using? I'm not seeing that error. I'm also wondering why the fix isn't to change the source to 11 to match the compiler.

arjantijms:
> Which JDK are you using? I'm not seeing that error. I'm also wondering why the fix isn't to change the source to 11 to match the compiler.

Interesting. The issue isn't actually the JDK versions matching, but the modular vs non-modular nature of the code and the link, where it sees the JDK as modular (even 8). But setting the source level to 11 does fixes the issue too.

I'm not 100% sure why it does tbh, but it does.

I'll update the change to source 11 then. Btw JDK used was OpenJDK 11.0.2.

arjantijms:
Btw, the tests I'm doing to build is using

`mvn -U -C -DskipTests -Ddoclint=none -Poss-release,staging  clean package install:install`

and

`mvn clean install`

On both JDK 11 and JDK 17.

PR #442, opened 2021-11-16
Add a module-info.java

### LINKED ISSUE #201: Java JPMS modules

arjantijms:
Fixes #201

Module name is `jakarta.servlet` (Per the naming [conventions](https://github.com/jakartaee/specification-committee/blob/master/names.adoc))
Exports the 4 API packages

Opens the 4 API packages as well to prevent errors like:

```
[ERROR] jakarta.servlet.http.CookieTest.testValue  Time elapsed: 0 s  <<< ERROR!

java.lang.reflect.InaccessibleObjectException: 
    Unable to make jakarta.servlet.http.CookieTest() accessible: 
        module jakarta.servlet does not "opens jakarta.servlet.http" to unnamed module @ffaa6af
```

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -249,9 +249,6 @@
                 <configuration>
                     <archive>
                         <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-                        <manifestEntries>
-                            <Automatic-Module-Name>jakarta.servlet</Automatic-Module-Name>
-                        </manifestEntries>
                     </archive>
                     <excludes>
                         <exclude>**/*.java</exclude>
--- a/api/src/main/java/module-info.java
+++ b/api/src/main/java/module-info.java
@@ -19,8 +19,5 @@ module jakarta.servlet {
     exports jakarta.servlet.descriptor;
     exports jakarta.servlet.http;
     
-    opens jakarta.servlet;
-    opens jakarta.servlet.annotation;
-    opens jakarta.servlet.descriptor;
-    opens jakarta.servlet.http;
+    opens jakarta.servlet.resources;
 }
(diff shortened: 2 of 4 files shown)


markt-asf:
I'm not sure opening those packages by default is what we want to do. Changing the Surefire Plugin to:
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.0.0-M5</version>
    <configuration>
        <argLine>--add-opens=jakarta.servlet/jakarta.servlet.http=ALL-UNNAMED</argLine>
    </configuration>
</plugin>
```
is perhaps a better approach.

I'm also not sure about `jakarta.servlet.resources`. I'm currently testing this locally to see what - if anything - is required for the schema to be accessible from the API JAR in a JPMS environment.

stuartwdouglas (changes requested):
Can we just make CookieTest and all its methods public instead?

markt-asf:
> Can we just make CookieTest and all its methods public instead?

stuartwdouglas:
The issue here though is that the test is private, and part of the same package, so JUnit can't load the test. Just make the test public.

markt-asf:
Confirmed. Making the test class and each test method public is sufficient to resolve the error. That solution works for me. (still need to fix resources)

stuartwdouglas (changes requested):
This will also need an entry in the changelog.

arjantijms:
There's seemingly a debate surrounding how to go about adding tests to the same Maven module in a modular Java world. The short of it, it's not entirely trivial, and there's a mismatch between Maven and Java modules. Maven distinguishes between regular and test dependencies, while Java modules does not.

Luckily the tests here are simple enough and don't have dependencies of their own.

arjantijms:
@stuartwdouglas @markt-asf See updated PR

markt-asf (approved):
LGTM

stuartwdouglas (approved):

PR #446, opened 2021-11-18
Fix #37 - update Cookie class for RFC 6265

### LINKED ISSUE #37: Update Cookie class and other specifications for RFC 6265
Currently the Cookie class defaults to supporting RFC 2019 cookies.
RFC2019 was obsoleted by RFC2965 in 2000, which in turn was obsoleted by RFC6265 in 2011

The latest RFC appears to be well supported by browsers (eg Google cookies often contain commas which are not allowed by 2019, but are by 6265).

markt-asf:
In summary:
- setVersion() is NO-OP, getVersion() is hard-coded to 0
- Restrictions on cookie name relaxed to 'must be non-null, non-empty token'

The reason I am proposing this for 6.0 as that I started to look at the TCK and noted that Tomcat's "pass the TCK" mode included various settings for use RFC 2109 rather than RFC 6265. I also recall a TCK challenge caused by the Javadoc continuing to refer to RFC 2109. It struck me that we really should try and get this into 6.0.

The changes turned out to be fairly minimal. Effectively disabling version should have minimal effect as RFC 2965 was never really implemented by any browser.

Reducing the restrictions on cookie names should safe as well.

### DIFF
--- a/api/src/main/java/jakarta/servlet/SessionCookieConfig.java
+++ b/api/src/main/java/jakarta/servlet/SessionCookieConfig.java
@@ -109,31 +109,33 @@ public interface SessionCookieConfig {
     public String getPath();
 
     /**
-     * Sets the comment that will be assigned to any session tracking cookies created on behalf of the application
-     * represented by the <tt>ServletContext</tt> from which this <tt>SessionCookieConfig</tt> was acquired.
-     *
+     * With the adoption of support for RFC 6265, this method should no longer be used.
      * <p>
-     * As a side effect of this call, the session tracking cookies will be marked with a <code>Version</code> attribute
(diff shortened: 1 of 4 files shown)


gregw (approved):
LGTM

markt-asf:
I'm intending to leave this PR open for further comment while I work on updating Tomcat to remove RFC 2109 support. I'm not expecting that to identify that further changes are required to this PR but you never know - and it gives folks more time to review the PR.

stuartwdouglas (approved):

markt-asf:
The forced pushed includes the following:
- rebase to master
- include 'Servlet' in any since values
- a change log entry

PR #450, opened 2021-11-22
Align Javadoc for isAsyncStarted() with specification document

markt-asf:


stuartwdouglas (approved):
LGTM, do you think this is worth including a changelog note for? We have not really talked about how big an update needs to be to be worth mentioning.

markt-asf:
Hmm. It isn't changing the intent of the spec or the expected behaviour of implementations so on that basis I'd lean towards no change log entry. However, if someone looked at that Javadoc in isolation then this clarification would look like a behavioural change so there are benefits of putting it in the change log to draw attention to it. I don't think a change log entry would do any harm. On balance, I think it is best to add one.

PR #456, opened 2022-01-08
Align isReady in ServletOutputStream with ServletInputStream

markt-asf:
The current wording is causing confusion:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65780

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletOutputStream.java
+++ b/api/src/main/java/jakarta/servlet/ServletOutputStream.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020 Oracle and/or its affiliates and others.
+ * Copyright (c) 1997, 2022 Oracle and/or its affiliates and others.
  * All rights reserved.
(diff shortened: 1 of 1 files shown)


martin-g (approved):

api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -272,10 +272,9 @@ public void println(double d) throws IOException {
     }
 
     /**
-     * This method can be used to determine if data can be written without blocking.
+     * Returns true if data can be written without blocking else returns false.
```
gregw, line 275: Probably we should also say something about the scheduling ramifications of a false return

markt-asf: +1 - for both `ServletInputStream` and `ServletOutputStream`. Do you have some text in mind? A new PR is probably the simplest way to proceed.

gregw: will try to do a PR soon.

PR #457, opened 2022-01-18
Update javadoc for isReady

gregw:
Make clear the scheduling implications of a false return from isReady()


stuartwdouglas (approved):

markt-asf (approved):
Thanks Greg

PR #477, opened 2022-11-14
Fix issue #325 - Clarify get[Int|Date]Header() for multiple headers

### LINKED ISSUE #325: Clarify HttpServletRequest.getDateHeader() when there are multiple field lines
Consider an HTTP message with two field lines for "If-Modified-Since", the first one a valid HTTP-date, the second one anything else.

Will "getDateHeader()" parse this?

(I believe it shouldn't, as RFC 7232 asks to ignore invalid dates in conditional header fields).

markt-asf:
Fixes #325 

This one is not clear-cut. There are a couple of different options for how we could handle this.

This PR implements the more backwards compatible approach which is to align with  `getHeader()` and state that if there are multiple headers with the same name, only the first is used.

Alternatives include:
- throw an exception if there is more than one header with the given name
- throw an exception if there is more than one header with the given name and the values are not identical (return the value if they are identical).
- something else

Given the potential for breakage if an exception is thrown, I think alignment with `getHeader()` is the way to go.

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021 Oracle and/or its affiliates and others.
+ * Copyright (c) 1997, 2022 Oracle and/or its affiliates and others.
  * All rights reserved.
  * Copyright 2004 The Apache Software Foundation
  *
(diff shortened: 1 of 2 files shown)


gregw (approved):
I think this is a good clarification for the getDateHeader method itself.  Note that a server is free to implement a stricter approach on it's actual conditional header handling.

stuartwdouglas (approved):

PR #479, opened 2022-11-17
Fix #59 - Add 'query_string' attribute to error dispatches

### LINKED ISSUE #59: Add 'query_string' request attribute to the Error dispatched requests
The servlet specification defines javax.servlet.forward.query_string and javax.servlet.include.query_string, but no javax.servlet.error.query_string.
If there is not specific reason why this attribute is excluded from the specification please add it for the next release.

markt-asf:
Fixes #59 

Mainly a note to self - need to update ErrorData in pages if this is merged.


stuartwdouglas (changes requested):
This should probably have a changelog entry.

markt-asf:
Agreed. I was trying to avoid merge conflicts with multiple PRs in flight and was planning on batching up the change log entries into a separate PR.

stuartwdouglas (approved):

Issue #473, opened 2022-11-02
TCK Challenge: isTrailerFields JavaDoc and default implementation inconsistent


mnriem:
<pre>
     * @implSpec The default implementation returns false.
     *
     * @return a boolean whether trailer fields are ready to read
     *
     * @since Servlet 4.0
     */
    default public boolean isTrailerFieldsReady() {
        return true;
    }
</pre>

Note the '@implSpec The default implementation returns false.' states it should return false whereas the actual default implementation returns true. Which is it?

markt-asf:
I removed the TCK challenge prefix since no TCK test was referenced in issue description.

I think the code is correct and the Javadoc needs to be updated. That would make the behaviour consistent with an implementation that did not implement trailer fields. `isTrailerFieldsReady()` would return `true` and a subsequent call to `getTrailerFields()` would return an empty Map.

mnriem:
@markt-asf I am asserting this is a valid TCK challenge. For the TCK test in question, 
see https://github.com/eclipse-ee4j/jakartaee-tck/blob/da6a34066dd04a78e6b1656c923c9c5de97af182/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/httpservletrequest40/Client.java#L295 for the failing TCK test

This is a valid TCK challenge so this test needs to be excluded because it fails on a spec / JavaDoc compliant implementation. Now for the next Servlet release this should be fixed in the JavaDoc and the test should be re-enabled. And then as part of the next Servlet release this should be called out as an incompatibility.

markt-asf:
I don't understand how this is failing since the rest of the Javadoc for that method describes how the method is expected to behave and it is this behaviour that the TCK test is testing. I don't see anything in the TCK that explicitly tests the default.

Further, I have run that TCK test on Tomcat 10.1.x with the default as both `true` and `false` and the test passes (as expected since a spec compliant implementation is going to have to override the method anyway).

I am currently leaning towards rejecting this challenge.

mnriem:
Again, if any implementation out there is implementing according to the specification or JavaDoc the default behavior requires to return false. You are saying that Tomcat 10.1.x passes this test. This simply means that your implementation is NOT using the as per JavaDoc prescribed default code path as any part of your implementation. It does not mean that the TCK challenge does not exist.

markt-asf:
Which version of which container using which Servlet API JAR fails this TCK test?

markt-asf:
[...]

PR #480, opened 2022-11-17
Fix Javadoc issue highlighted by #473

markt-asf:
Align Javadoc with current behaviour.

The default current behaviour is consistent with an implementation that does not implement trailer fields. isTrailerFieldsReady() returns true and a subsequent call to getTrailerFields() returns an empty Map.


arjantijms (approved):

stuartwdouglas:
Should this have had a changelog entry? It only occurred to me after I hit merge but IMHO this is basically just a typo fix so from a users point of view nothing has changed.

markt-asf:
With multiple PRs in flight I wanted to avoid merge conflicts. I was planning on batching up the change log entries and adding them in a separate PR.

PR #482, opened 2022-11-18
Fix #95 - clarify the HTTPS support is required

### LINKED ISSUE #95: Require that TLS is supported
Section 1.2 states:

> All servlet containers must support HTTP as a protocol for requests and responses, but additional request/response-based protocols such as HTTPS (HTTP over SSL) may be supported.

With the recent publishing of RFC 7258 / BCP 188 [http://www.rfc-editor.org/rfc/rfc7258.txt](http://www.rfc-editor.org/rfc/rfc7258.txt) TLS is encouraged to be the default for servers.

We will need to revisit the above text in light of BCP 188.

markt-asf:


markt-asf:
On the basis this should be complete uncontroversial, I am going to merge this change.

PR #483, opened 2022-11-22
Fix #275 - containers can use relative URLS for redirects

### LINKED ISSUE #275: Clarify servlet response sendRedirect()
Hi,

https://jakarta.ee/specifications/servlet/4.0/apidocs/javax/servlet/http/HttpServletResponse.html#sendRedirect-java.lang.String-

The API doc states that "the servlet container must convert the relative URL to an absolute URL before sending the response to the client."

Currently, different server returns different Location header.

Tomcat returns a relative path, for example : Location: /reDirectPath

Jetty, WebSphere Liberty return an absolute URL, for example: Location: http://x.x.x.x:9080/reDirectPath

Can you please clarify the expecting behavior? Perhaps, clearly state it in the next servlet specification/doc.

Thanks,

markt-asf:
Doesn't mandate any new behaviour, just opens the door for containers that want to to use relative redirects while remaining specification compliant.

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
@@ -147,12 +147,21 @@ public interface HttpServletResponse extends ServletResponse {
     /**
      * Sends a temporary redirect response to the client using the specified redirect location URL and clears the buffer.
      * The buffer will be replaced with the data set by this method. Calling this method sets the status code to
-     * {@link #SC_FOUND} 302 (Found). This method can accept relative URLs;the servlet container must convert the relative
-     * URL to an absolute URL before sending the response to the client. If the location is relative without a leading '/'
-     * the container interprets it as relative to the current request URI. If the location is relative with a leading '/'
-     * the container interprets it as relative to the servlet container root. If the location is relative with two leading
-     * '/' the container interprets it as a network-path reference (see <a href="http://www.ietf.org/rfc/rfc3986.txt"> RFC
-     * 3986: Uniform Resource Identifier (URI): Generic Syntax</a>, section 4.2 &quot;Relative Reference&quot;).
+     * {@link #SC_FOUND} 302 (Found).
+     * <p>
+     * This method accepts both relative and absolute URLs. Absolute URLs passed to this method are used as provided as the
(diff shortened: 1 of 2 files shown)


markt-asf:
Ping. It has been a week. Any comments before this is merged?

api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
```diff
@@ -147,12 +147,22 @@ public interface HttpServletResponse extends ServletResponse {
...
-     * the container interprets it as relative to the current request URI. If the location is relative with a leading '/'
-     * the container interprets it as relative to the servlet container root. If the location is relative with two leading
-     * '/' the container interprets it as a network-path reference (see <a href="http://www.ietf.org/rfc/rfc3986.txt"> RFC
-     * 3986: Uniform Resource Identifier (URI): Generic Syntax</a>, section 4.2 &quot;Relative Reference&quot;).
+     * {@link #SC_FOUND} 302 (Found).
+     * <p>
+     * This method accepts both relative and absolute URLs. Absolute URLs passed to this method are used as provided as the
+     * redirect location URL . A servlet container may provide a container specific option that controls whether relative
+     * URLs passed to this method are converted to absolute URLs or used as provided for the redirect location URL. If no
+     * such option is provided then the servlet container must always convert a relative URL passed to this method to an
```
gregw, line 155: I don't think this is needed.  A container may provide an option and that option might be hard wired true or false.. or it might be switched by changing code and re-releasing.... but it is always an option.

I think it is enough to say that it is container specific.

gregw (approved):
OK other than a niggle

api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
markt-asf: OK. Let me see if I can re-word that to simplify it a bit.

markt-asf: How about this:

This method accepts both relative and absolute URLs. Absolute URLs passed to this method are used as provided as the redirect location URL. Relative URLs are converted to absolute URLs unless a container specific feature/option is provided that controls whether relative URLs passed to this method are converted to absolute URLs or used as provided for the redirect location URL. If converting a relative URL to an absolute URL then:

PR #485, opened 2022-11-25
Fix issue-484 Add read/write via ByteBuffer

markt-asf:
Could be applied as-is but more likely to be a starting point for discussion.

The default implementations could be more efficient but I opted for clarity over efficiency given that containers should be overriding these methods anyway.

One obvious question - to we want to add scatter/gather read/write?

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletInputStream.java
+++ b/api/src/main/java/jakarta/servlet/ServletInputStream.java
@@ -20,9 +20,11 @@ package jakarta.servlet;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.Objects;
 
 /**
- * 
+ *
  * Provides an input stream for reading binary data from a client request, including an efficient <code>readLine</code>
  * method for reading data one line at a time. With some protocols, such as HTTP POST and PUT, a
  * <code>ServletInputStream</code> object can be used to read data sent from the client.
@@ -51,14 +53,75 @@ public abstract class ServletInputStream extends InputStream {
     }
 
     /**
+     * Reads from the input stream into the given buffer.
+     * <p>
+     * If the input stream is in non-blocking mode, before each invocation of this method {@link isReady()} must be called
+     * and must return {@code true} or the {@link ReadListener#onDataAvailable()} call back must indicate that data is
+     * available to read else an {@link IllegalStateException} must be thrown.
+     * <p>
+     * Otherwise, if this method is called when {@code buffer} has no space remaining, the method returns {@code 0}
+     * immediately and {@code buffer} is unchanged.
+     * <p>
+     * If the input stream is in blocking mode and {@code buffer} has space remaining, this method blocks until at least one
+     * byte has been read, end of stream is reached or an exception is thrown.
+     * <p>
+     * Returns the number of bytes read or {@code -1} if the end of stream is reached without reading any data.
+     * <p>
+     * When the method returns, and if data has been read, the buffer's position will be unchanged from the value when
+     * passed to this method and the limit will be the position incremented by the number of bytes read.
+     * <p>
+     * Subclasses are strongly encouraged to override this method and provide a more efficient implementation.
      *
+     * @param buffer The buffer into which the data is read.
+     *
+     * @return The number of bytes read or {@code -1} if the end of the stream has been reached.
+     *
+     * @exception IllegalStateException If the input stream is in non-blocking mode and this method is called without first
+     * calling {@link #isReady()} and that method has returned {@code true} or {@link ReadListener#onDataAvailable()} has
+     * not signalled that data is available to read.
+     * 
+     * @exception IOException If data cannot be read for any reason other than the end of stream being reached, the input
+     * stream has been closed or if some other I/O error occurs.
+     * 
+     * @exception NullPointerException If buffer is null.
+     * 
+     * @since Servlet 6.1
+     */
+    public int read(ByteBuffer buffer) throws IOException {
+        Objects.requireNonNull(buffer);
+
+        if (!isReady()) {
+            throw new IllegalStateException();
+        }
+
+        if (buffer.remaining() == 0) {
+            return 0;
+        }
+
+        byte[] b = new byte[buffer.remaining()];
+
+        int result = read(b);
+        if (result == -1) {
+            return -1;
+        }
+
+        int position = buffer.position();
+
+        buffer.put(b, 0, result);
+
+        buffer.position(position);
+        buffer.limit(position + result);
+
+        return result;
+    }
+
+    /**
      * Reads the input stream, one line at a time. Starting at an offset, reads bytes into an array, until it reads a
      * certain number of bytes or reaches a newline character, which it reads into the array as well.
-     *
      * <p>
-     * This method returns -1 if it reaches the end of the input stream before reading the maximum number of bytes.
-     *
-     *
+     * This method returns {@code -1} if it reaches the end of the input stream before reading the maximum number of bytes.
+     * <p>
+     * This method may only be used when the input stream is in blocking mode.
      *
      * @param b an array of bytes into which data is read
      *
@@ -68,8 +131,9 @@ public abstract class ServletInputStream extends InputStream {
      *
      * @return an integer specifying the actual number of bytes read, or -1 if the end of the stream is reached
      *
+     * @exception IllegalStateException If this method is called when the input stream is in non-blocking mode.
+     * 
      * @exception IOException if an input or output exception has occurred
-     *
      */
     public int readLine(byte[] b, int off, int len) throws IOException {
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -51,6 +53,37 @@ public abstract class ServletOutputStream extends OutputStream {
...
+     * Subclasses are strongly encouraged to override this method and provide a more efficient implementation.
+     *
+     * @param buffer The buffer from which the data is written
+     *
+     * @exception IOException If the output stream has been closed or if some other I/O error occurs.
+     *
+     * @exception NullPointerException If buffer is null
+     */
+    public void write(ByteBuffer buffer) throws IOException {
+        Objects.nonNull(buffer);
```
martin-g, line 74: I think you wanted to use `requireNonNull()` instead.

martin-g (approved):

api/src/main/java/jakarta/servlet/ServletOutputStream.java
markt-asf: That would be rather more useful wouldn't it ;).
Tx for spotting that. I'll get it fixed shortly.

api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -51,6 +53,37 @@ public abstract class ServletOutputStream extends OutputStream {
...
 
+    /**
+     * Writes from the given buffer to the output stream. The method blocks until all the remaining data in the buffer has
+     * been written.
+     * <p>
+     * If this method is called when the buffer has no data remaining, the method returns immediately.
+     * <p>
+     * When the method returns, and if data has been written, the buffer's limit will be unchanged from the value when
+     * passed to this method and the position will be the same as the limit.
+     * <p>
```
gregw, line 64: This may not be true for asynchronous writes.  It is true for asynchronous write once this method has returned AND `isReady()` has returned true.

gregw: More over, callers must not modify the position, limit nor the underlying buffer content until `isReady()` has returned true.

gregw (changes requested):

api/src/main/java/jakarta/servlet/ServletOutputStream.java
markt-asf: Good points. I based the text on the existing [In|Out]putStream text which doesn't address non-blocking. I'll update the descriptions.

markt-asf:
I've updated the text.

One further question occurred to me. The implement can't (easily) perform the check that would throw an ISE if the stream isn't ready for a non-blocking read/write. The options I can see are:

1. ignore the missing check
1. drop the implementation and make the method abstract
1. add additional plumbing that enables us to determine if the stream is in non-blocking mode

Ignoring the check seems wrong. The additional plumbing might not be possible since any concrete implementation we add for setting the listener (so we can detect if we are in non-blocking mode) will have no effect as the container's implementation will override it. I think that leaves making these new methods abstract.

gregw:
@markt-asf Can't the implementations just call `isReady()` and throw ISE on a false return?      Not very efficient, but then it's not meant to be.

markt-asf:
That would work for non-blocking mode. In blocking mode, `isReady()` could return false since it is defined (in the Javadoc) as "Returns true if data can be read without blocking else returns false.". That would trigger the ISE when we don't want one.

If we redefined `isReady()` so it always returned `true` when in blocking mode then yes, we could simply check `isReady()`. That seems reasonable to me since there is no reason apps should be calling `isReady()` when in blocking mode.

There is a risk that apps are doing something odd and calling `isReady()` when still in blocking mode but it seems low enough to me that changing the definition is acceptable. Thoughts?

gregw:
I think returning true for isReady in blocking mode is fine.
Perhaps saying that a true return means that it is allowed to call the API, rather than saying you can call "without blocking."... which kind of implies that you can call it, but it would block which is not the case.

api/src/main/java/jakarta/servlet/ServletInputStream.java
```diff
@@ -50,6 +52,63 @@ public abstract class ServletInputStream extends InputStream {
     protected ServletInputStream() {
     }
 
+    /**
+     * Reads from the input stream into the given buffer. Returns the number of bytes read or {@code -1} if the end of
+     * stream is reached without reading any data.
+     * <p>
+     * If this method is called when {@code buffer} has no space remaining, the method returns {@code 0} immediately.
```
stuartwdouglas, line 59: What do you think about making this an error instead? The current implementation might make it a bit easier for buggy application code to get stuck in an infinite loop.

markt-asf: Returning zero is consistent with `InputStream.read(byte[])` and with `SocketChannel.read(ByteBuffer)`. I understand the motivation but my preference is for consistency.

markt-asf:
@gregw I'll work on an update that defines `isReady()` returns true in the blocking case.

markt-asf:
Updated with `isReady()` changes and added throwing ISE to read/write methods.

api/src/main/java/jakarta/servlet/ServletInputStream.java
gregw: Does this immediate return of 0 happen even if the method is called in non blocking mode with isReady false?
I think that should still be an error. Maybe leave out "immediately"?

markt-asf: I agree that should still be an ISE and that is what the provided implementation does. I'll see what I can do to clarify the wording.

markt-asf:
Language updated to clarify various ambiguities.

api/src/main/java/jakarta/servlet/ServletInputStream.java
```diff
@@ -50,6 +52,69 @@ public abstract class ServletInputStream extends InputStream {
     protected ServletInputStream() {
     }
 
+    /**
+     * Reads from the input stream into the given buffer.
+     * <p>
+     * If the input stream is in non-blocking mode, before each invocation of this method {@link isReady()} must be called
+     * and must return {@code true} or the {@link ReadListener#onDataAvailable()} call back must indicate that data is
```
stuartwdouglas, line 59: I always interpreted the spec as always requiring `isReady()` to return `true` before you can call `read`, even in an `onDataAvailible` callback. Seeing this made me re-read the relevant sections. The `isReady` section says:

```
The `isReady` method returns `true` if data can be read without blocking. If no data can be
read without blocking it returns `false`. If `isReady` returns `false` it
is illegal to call the read method and an `IllegalStateException` MUST
be thrown.
```

To me that implies that once `isReady` returns false `read` must throw an ISE until it has been called again and returns `true`, even if `onDataAvailable` has been called.

IMHO it makes sense to allow for `read` to be called in `onDataAvailable` before the `isReady` call, but as written I don't think the spec allows it. 

If we are going to allow it we should clarify this on the `isReady` docs as well.

stuartwdouglas (changes requested):

api/src/main/java/jakarta/servlet/ServletInputStream.java
gregw: Wow, nothing is ever simple!!

So how about this for the input `isReady()` javadoc:

> The `isReady` method returns `true` if it is allowable to call the `read` method. In blocking mode, `isReady` will always return `true`, but a subsequent call to `read` may block awaiting data. In non-blocking mode the `isReady` method may return `false`, in which case it is illegal to call the read method and an `IllegalStateException` MUST be thrown. When `onDataAvailable` is called, a call to `isReady` that returned `true` is implicit.

and for the out `isReady()` javadoc:

The `isReady` method returns `true` if it is allowable to call any method that may write data (e.g. `write`, `print` or `flush`).  In blocking mode, `isReady` will always return `true`, but a subsequent call to method that write data may block. In non-blocking mod the `isReady` method may return false, in which case it is illegal to call a method that writes data and an `IllegalStateException` MUST be thrown. When `onWriteAvailable` is called, a call to `isReady` that returned `true` is implicit.

markt-asf: For output, barring fixing a couple of typos, that looks good.
For input, you changed the language from "a read method" to "the `read` method" which begs the question which `read` method? Any of the `read` methods are OK for use in non-blocking mode. The other methods that read data are not. If we changed the text to "a `read` method" that could be sufficient but do we want to consider making calls to the other methods that read data throw an ISE in non-blocking mode or leave the behaviour as undefined?

stuartwdouglas: If you are talking about methods like `readAllBytes` think we should mandate an ISE. This is the sort of thing that will work fine in your tests where everything was buffered with your small amount of test data, then fail in production.

gregw: I'm fine with "a read method" rather than "the read method".

The only doubt I have, for both read/write is if `close` is a method that reads/writes.    I think on input close is not a read. On output close can be a write.. but I also think we treat it specially and allow it to be called at anytime (even though that makes the implementation a lot more complex if there is a pending write in progress).

stuartwdouglas: I think it can be called anytime. If we were going to specify it I would say something like:

`If close is called when the stream is in async mode it will immediately return even if the stream contains buffered data that is yet to be written to client. The stream will write this data out in the background, and then close the response. If this process fails the onError method will be invoked as normal.`

markt-asf:
OK. I have updated the Javadoc to take account of most of the recent discussion.

I haven't yet changed the Javadoc for `ServletOutputStream.close()`. Tomcat does something different and I want to think about this use case further first.

markt-asf:
Close text added with just minor changes for consistency.

api/src/main/java/jakarta/servlet/ServletOutputStream.java
```diff
@@ -304,4 +362,16 @@ public void println(double d) throws IOException {
      */
     public abstract void setWriteListener(WriteListener writeListener);
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * If this method is called when the output stream is in non-blocking mode it will immediately return even if the stream
+     * contains buffered data that is yet to be written to client. The container will write this data out in the background,
+     * and then close the response. If this process fails the {@link WriteListener#onError(Throwable)} method will be
+     * invoked as normal.
```
gregw, line 371: Is it "and then close the response"?    I think the response is functionally closed when the method returns (write cannot be called again), but that the return does not guarantee that the data has been written.
```suggestion
     * If this method is called when the output stream is in non-blocking mode it will immediately return with the stream effective close, even if the stream
     * contains buffered data that is yet to be written to client. The container will write this data out in the background. If this process fails the {@link WriteListener#onError(Throwable)} method will be
     * invoked as normal.
```

stuartwdouglas: `with the stream effective close` should be `with the stream effectively closed`

markt-asf:
Wording updated.

gregw (approved):

stuartwdouglas (approved):

PR #487, opened 2022-12-02
Fix issue 453

### LINKED ISSUE #453: Clarify normalization of methods taking paths
After #424 for #18, we need to further clarify how methods that take paths will treat non-normal and/or encoded characters.  

An example of an issue could be an application that take pathInfo for a servlet mapped to `/` and uses that to request a resource or request dispatcher.  If a URI like `/foo/%252e%252e/WEB-INF/web.xml` is received, then that will decode to a pathInfo of `/foo/%25%25/WEB-INF/web.xml`.    How should that arg be treated by thos methods?   Should it be rejected because the method doesn't accept encoded paths?  Or should it be decoded and normalized to `/WEB-INF/web.xml`, but then rejected because it is not the canonical path?  

Note that `getRequestDispatcher` allows for relative paths within the docroot, so asking for `../foo/file.txt` should be allowed, even though it is not canonical.  But should `%2e%2e/foo/file.txt` be allowed?

The `getResource` method has the following warning:
>   This method bypasses both implicit (no direct access to WEB-INF or META-INF) and explicit (defined by the web
     application) security constraints. Care should be taken both when constructing the path (e.g. avoid unsanitized user
     provided data) and when using the result not to create a security vulnerability in the application.

Methods that need clarification are: 

- ServletContext#getRequestDispatcher(java.lang.String)
- ServletContent#getRealPath(java.lang.String)
- ServletContent#getResource(java.lang.String)
- ServletContent#getResourcePaths(java.lang.String)
- ServletContent#getResourceAsStream(java.lang.String)
- ServletRequest#getRequestDispatcher(java.lang.String)

markt-asf:
This is intended to fix #453

The first commit should be uncontroversial.

I suspect the second commit will require some discussion but, logically, I don't currently see any viable alternative. My reasoning for this is explained in https://github.com/jakartaee/servlet/issues/453#issuecomment-982404860

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -212,6 +212,11 @@ public interface ServletContext {
      * application) security constraints. Care should be taken both when constructing the path (e.g. avoid unsanitized user
      * provided data) and when using the result not to create a security vulnerability in the application.
      * 
+     * <p>
+     * The provided {@code path} parameter is canonicalized as per <a href=
+     * "https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization">Servlet 6.0,
+     * 3.5.2</a> before being used to match resources.
+     * 
      * @param path the partial path used to match the resources, which must start with a <tt>/</tt>
      * @return a Set containing the directory listing, or null if there are no resources in the web application whose path
      * begins with the supplied path.
@@ -257,6 +262,11 @@ public interface ServletContext {
      * application) security constraints. Care should be taken both when constructing the path (e.g. avoid unsanitized user
      * provided data) and when using the result not to create a security vulnerability in the application.
      *
+     * <p>
+     * The provided {@code path} parameter is canonicalized as per <a href=
+     * "https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization">Servlet 6.0,
+     * 3.5.2</a> before being used to match resources.
+     * 
      * @param path a <code>String</code> specifying the path to the resource
      *
      * @return the resource located at the named path, or <code>null</code> if there is no resource at that path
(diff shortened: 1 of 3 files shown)


markt-asf:
Forgot to update the copyright.

markt-asf:
I think this is worth an explicit review because of the second commit.

api/src/main/java/jakarta/servlet/ServletContext.java
```diff
@@ -212,6 +212,10 @@ public interface ServletContext {
      * application) security constraints. Care should be taken both when constructing the path (e.g. avoid unsanitized user
      * provided data) and when using the result not to create a security vulnerability in the application.
      * 
+     * <p>
+     * The provided {@code path} parameter is canonicalized as per section 3.5.2 of the specification before being used to
```
stuartwdouglas, line 216: I don't really like this. As far as I can see this would actually set a new precedent, as none of the other javadoc we have explicitly references a section of the spec (I could be wrong, but it seemed like all other references were to RFC's rather than the Servlet spec).

If we are going to do this I think we need to explicitly specify the version of the spec, as at some point the section number will change and we may forget to update the javadoc.

stuartwdouglas (changes requested):

api/src/main/java/jakarta/servlet/ServletContext.java
stuartwdouglas: Turns out I was wrong, there is this:

https://github.com/jakartaee/servlet/blob/100fd4e8d817ca2ca1f9e892c80e1c29b07e2aa0/api/src/main/java/jakarta/servlet/http/HttpServletRequest.java#L257

This does not even specify which spec, let alone the version :-)

I really think that over time this will likely become outdated and rot, as these number can change if additional items are added, and it would be very easy to miss that they need to be updated, although I don't really have a good solution.

gregw: We should probably come up with a very standard javadoc style for referencing a specific spec document and the section within it, so that we can easily find and update in future.

markt-asf: How about this (with hyperlink to the right section in the HTML version of the spec):
[Servlet 6.0, 3.5.2](https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization)
We might need to tweak to working around the reference so that style works everywhere.

gregw: That works for me

stuartwdouglas: Works for me as well.

markt-asf:
Servlet specification references updated to agreed format and PR rebased.

stuartwdouglas (approved):

PR #491, opened 2022-12-13
Fix #415 - add overloaded setCharacterEncoding() supporting Charset

### LINKED ISSUE #415: Add overload to setCharacterEncoding supporting Charset
In JDK, a lot of methods that supported String as charset got overloads to support the `java.nio.charset.Charset`. It makes the lookup faster because it doesn't require a parsing of the String to know if it supported.

One example of this would be the methods `ServletRequest.setCharacterEncoding` and `ServletResponse.setCharacterEncoding`.

Maybe add these overloads as default methods.

markt-asf:


markt-asf:
It has been a few weeks but those weeks were over the holiday period. This PR should be uncontroversial so I'll give it another week or so before merging.

PR #496, opened 2023-01-16
Fix #273 - clarify behaviour around non-blocking writes

### LINKED ISSUE #273: Clarify behaviour after ServletOutputStream.setWriteListener
The current javadoc for [`ServletOutputStream.setWriteListener`](https://jakarta.ee/specifications/servlet/4.0/apidocs/javax/servlet/ServletOutputStream.html#setWriteListener-javax.servlet.WriteListener-) is minimal and does not well define the async behaviour, specifically in regards to which methods are affected.   The 4.0 servlet specification says:
```
boolean isReady(). This method returns true if a write to the ServletOutputStream will succeed, 
otherwise it will return false. If this method returns true, a write operation can be performed on 
the ServletOutputStream. If no further data can be written to the ServletOutputStream. then this 
method will return false till the underlying data is flushed at which point the container will invoke 
the onWritePossible method of the WriteListener. A subsequent call to this method will return true.
```

As the specification is no longer available for clarifications, this behaviour needs to be codified in the javadoc and clarifications made there.
Specifically it would be desirable to define what is meant by "a write operation", as writes may be performed by methods like 
`ServletOutputStream.close()`, `ServletResponse.flushBuffers()`, `AsyncContext.complete()`, `ServletResponse.sendError(...)`, etc.

See also #256

markt-asf:
This turned out to be a much smaller change than I was expecting - primarily because applications still need to call `complete()` or `dispatch()` after they call `ServletOutputSream.close()`.

### DIFF
--- a/api/src/main/java/jakarta/servlet/AsyncContext.java
+++ b/api/src/main/java/jakarta/servlet/AsyncContext.java
@@ -154,12 +154,22 @@ public interface AsyncContext {
      *
      * <p>
      * This method returns immediately after passing the request and response objects to a container managed thread, on
-     * which the dispatch operation will be performed. If this method is called before the container-initiated dispatch that
-     * called <tt>startAsync</tt> has returned to the container, the dispatch operation will be delayed until after the
-     * container-initiated dispatch has returned to the container. If the output stream is in non-blocking mode when this
-     * method is called, the output stream will be closed as described by {@code ServletOutputStream#close} and the dispatch
-     * operation will be delayed until after the non-blocking write has completed.
-     *
+     * which the dispatch operation will be performed. If this method is called during a container-initiated dispatch, the
+     * dispatch operation will be delayed until after the container-initiated dispatch has returned to the container. <br>
+     * Note: Container initiated dispatches include calls to:
+     * <ul>
+     * <li>{@link Servlet#service(ServletRequest, ServletResponse)} where {@link ServletRequest#startAsync()} or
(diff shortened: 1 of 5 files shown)


markt-asf:
There haven't been any comments on this in the week or so since I opened it. I'll give it a couple more days and then merge it.

gregw (approved):
Your changes are good.

However, I think we might need to do a bit more as the following form of words used in a few places, is not exactly correct:
```
     *  If this method is called before the container-initiated dispatch that called
     * <tt>startAsync</tt> has returned to the container, then the call will not take effect (and any invocations of
     * {@link AsyncListener#onComplete(AsyncEvent)} will be delayed) until after the container-initiated dispatch has
     * returned to the container.
```

I think that all dispatches are serialized, so any container-initiated dispatch (regardless if it is the one that called `startAsync`) must return before another dispatch.  Specifically if you call `isReady()` in `onWritePossible()` and get a false return, OWP will not be called again until the first call returns.

So perhaps just 

```
     * If this method is called before a container-initiated dispatch 
     * has returned to the container, then the call will not take effect until after the 
     * container-initiated dispatch has returned to the container.
```

markt-asf:
I agree with your views regarding the serialization of dispatches and that further clarifications would be helpful. I've made some additional changes to the Javadoc. WDYT?

markt-asf:
Unless there are objections, I'll merge this later this week.

stuartwdouglas (approved):

PR #502, opened 2023-02-13
Patterns normalization

tbw777:


markt-asf:
Wrong project. You want:
https://github.com/eclipse-ee4j/jakartaee-schemas

Note: I suspect that changes to historic schema that have been released as part of a specification will be rejected. Changes to the Jakarta EE 11 schemas should be fine.

PR #503, opened 2023-02-21
Fix #74 - provide more control over redirects

### LINKED ISSUE #74: HttpServletResponse#sendRedirect(String) could use some improvement; needs 301/302/303 support; needs way to set request body
Currently, <tt>HttpServletResponse#sendRedirect(String)</tt> specifies the following hard-wired behavior that cannot be changed:

*   Send the redirect using status code <tt>302 Found</tt> and the <tt>Location</tt> header.
*   Clear the buffer and set the response body to a short hypertext note with a hyperlink to the new URI.

I have three issues with this behavior that I summarize below, and at the bottom I propose a solution:

#### <a name="Issue1"></a>Issue 1

[RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), says the following about <tt>302 Found</tt>:

> If the 302 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued.
> 
> Note: RFC 1945 and RFC 2068 specify that the client is not allowed to change the method on the redirected request. However, most existing user agent implementations treat 302 as if it were a 303 response, performing a GET on the Location field-value regardless of the original request method.

The two keys here are 1) that the spec says browsers MUST NOT automatically redirect the request and 2) that there could be some browsers that send the second request with the original method (<tt>POST</tt>, <tt>PUT</tt>, etc.).

As an application not concerned with legacy HTTP/1.0 user-agents and more concerned with ensuring that user agents do not prompt the user for permission to redirect and do not redirect using a <tt>POST</tt> or a <tt>PUT</tt>, etc., I prefer to use the <tt>303 See Other</tt> response as opposed to the <tt>302 Found</tt> response:

> The response to the request can be found under a different URI and SHOULD be retrieved using a GET method on that resource. This method exists primarily to allow the output of a POST-activated script to redirect the user agent to a selected resource. ... The 303 response MUST NOT be cached, but the response to the second (redirected) request might be cacheable.

#### <a name="Issue2"></a>Issue 2

As an application that has recently been refactored, I want to be able to easily send a permanent redirect to users for old URLs that have been removed and should be changed. This can be achieved with the <tt>301 Moved Permanently</tt> response:

> The requested resource has been assigned a new permanent URI and any future references to this resource SHOULD use one of the returned URIs. Clients with link editing capabilities ought to automatically re-link references to the Request-URI to one or more of the new references returned by the server, where possible. ... The new permanent URI SHOULD be given by the Location field in the response. Unless the request method was HEAD, the entity of the response SHOULD contain a short hypertext note with a hyperlink to the new URI(s). ... If the 301 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued.

#### <a name="Issue3"></a>Issue 3

The exact wording of the short hypertext response included with the current <tt>302 Found</tt> responses is not consistent across containers. As an application developer that cares about branding and consistency, I would like to be able to customize this text within my code and do so in a way that is consistent across all containers.

#### <a name="Recommendation"></a>Recommendation

I believe the following additions to the Servlet API will greatly improve the current <tt>HttpServletResponse#sendRedirect(String)</tt> and satisfy the needs outlined here.

```
**New HttpRedirectType Enum**public enum HttpRedirectType {
    /*
     * Indicates that the {@code 301 Moved Permanently} status should be used.
     * Could alternatively be called MOVED_PERMANENTLY.
     */
    PERMANENT,

    /*
     * Indicates that the {@code 302 Found} status should be used.
     * Could alternatively be called FOUND.
     */
    TEMPORARY_SAME_METHOD,

    /*
     * Indicates that the {@code 303 See Other} status should be used.
     * Could alternatively be called SEE_OTHER.
     */
    TEMPORARY_USE_GET;
}
```
```
**Additions to HttpServletResponse**public interface HttpServletResponse {
...
    void sendRedirect(String location, HttpRedirectType type);

    void sendRedirect(String location, boolean replaceBuffer);

    void sendRedirect(String location, HttpRedirectType type, boolean replaceBuffer);
...
}
```

*   If <tt>replaceBuffer</tt> is <tt>true</tt>, the container should clear the buffer and replace it with the data set by this method. If <tt>replaceBuffer</tt> is <tt>false</tt>, the container should send the existing buffer and should not add to it (even if it is empty) or remove from it.
*   The existing <tt>sendRedirect(String)</tt> should call <tt>sendRedirect(location, HttpRedirectType.TEMPORARY_SAME_METHOD, true)</tt> to preserve current behavior.
*   The new <tt>sendRedirect(String, HttpRedirectType)</tt> should call <tt>sendRedirect(location, type, true)</tt> to use the default response body.
*   The new <tt>sendRedirect(String, boolean)</tt> should call <tt>sendRedirect(location, HttpRedirectType.TEMPORARY_SAME_METHOD, replaceBuffer)</tt> to use the default <tt>302 Found</tt> response.

markt-asf:


api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
```diff
@@ -145,9 +145,57 @@ public interface HttpServletResponse extends ServletResponse {
     public void sendError(int sc) throws IOException;
 
     /**
-     * Sends a temporary redirect response to the client using the specified redirect location URL and clears the buffer.
-     * The buffer will be replaced with the data set by this method. Calling this method sets the status code to
+     * Sends a temporary redirect response to the client using the specified redirect location URL using the status code
+     * {@link #SC_FOUND} 302 (Found) and clears the buffer. The buffer will be replaced with the data set by this method.
+     *
+     * @param location the redirect location URL
```
gregw, line 151: Can we mention that the location can be absolute or relative here in the javadoc (and not in the intro to the full sendRedirect method). 

We should probably also list he IllegalArgumentException that the spec says should be used if a relative URI can not be converted.

gregw (commented):
I think there are a couple of other things we need to clarify in the javadoc:
 + do these methods commit the response?
 + can they be called from an include?
 + we should maybe point out that if the request has a body content, then it still needs to be consumed if the connection is to be persisted? or is that too much detail?

api/src/main/java/jakarta/servlet/http/HttpServletResponse.java
markt-asf: Agreed. I'll update the PR.

markt-asf:
Any more feedback on the updated PR? I'll leave it a couple of days and then merge unless there are objections.

PR #504, opened 2023-02-23
Clarify expected behaviour for CONNECT

markt-asf:


gregw (approved):

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -332,7 +332,10 @@ supported by the servlet. The `doTrace` method generates a response
 containing all instances of the headers sent in the `TRACE` request.
 
 The `CONNECT` method is not supported because it applies to proxies and
-the Jakarta Servlet API is targeted at endpoints.
+the Jakarta Servlet API is targeted at endpoints. Containers must not pass
```
stuartwdouglas, line 335: Are we sure we want to completely ban this? Maybe change it to 'are not required to' rather than 'must not' and note that the behaviour of the various request/response methods is implementation dependent.

gregw: I'm not sure what the value of saying that it may be passed to a filter/servlet but then not defining what the methods do in that circumstance.

I don't think a container can use filters/servlets to implement CONNECT because the input//output stream semantic is just not there.  Perhaps it could be tied somehow to the upgrade mechanism, but I just don't see the demand.
If a container wants to do something special, then it can ignore the "must not" in the knowledge that it is outside of the specification and no other filter/servlet should be expected to work if it sees such a request.

stuartwdouglas: What do you mean by 'the input//output stream semantic is just not there'. The servlet can just use the standard input/output streams, with it being up to the container to ensure that the correct HTTP response code is sent.

gregw: If you tried to implement a TCP/IP tunnel in a servlet, then you'd need to be able to read/write bytes to/from the connection.  But the input stream would be at EOF because there CONNECT has no body, and after you send the 200 response to initiate the tunnel, then you can try writing data, but it will be framed, probably as chunks, thus stopping you being able to send the arbitrary bytes that such a tunnel requires.  It just can't be done in a servlet, not without breaking all sorts of other spec requirements.

markt-asf: How about this text:
> The `CONNECT` method applies to proxies whereas the Jakarta Servlet API is
targeted at endpoints so, by default, Containers must reject such requests with
an `SC_NOT_IMPLEMENTED` (501) response and not pass the request to a Filter or
Servlet. Containers may provide container specific functionality to handle
`CONNECT` requests. If that functionality includes passing `CONNECT` requests to
a Filter or Servlet, then the container should define the expected behaviour of
the Servlet API for such requests in that container.

It also means we can add a test to the TCK since the default behaviour is defeined.

stuartwdouglas: For some reason I was thinking this would currently work in Undertow, but it turns out I was wrong. That said you could still make it work by treating CONNECT requests specially in the underlying Servlet code, but I don't think we would ever do that.

I like that version of the text.

PR #505, opened 2023-02-24
Fix #469 - Remove sensitive header on TRACE method

### LINKED ISSUE #469: AS per RFC 9110 (and earlier) HTTp TRACE response MUST not include sensitive headers
https://www.rfc-editor.org/rfc/rfc9110.html#name-trace

The requirement was added in RFC 7231. It is not present in RFC 2616.

dsandrade0:
This commit

### DIFF
--- a/api/src/main/java/jakarta/servlet/http/HttpServlet.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServlet.java
@@ -31,7 +31,9 @@ import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
 import java.text.MessageFormat;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.List;
 import java.util.ResourceBundle;
 
 /**
@@ -78,6 +80,7 @@ public abstract class HttpServlet extends GenericServlet {
 
     private static final String HEADER_IFMODSINCE = "If-Modified-Since";
     private static final String HEADER_LASTMOD = "Last-Modified";
+    private static final List<String> HEADER_SENSITIVE = Arrays.asList("Authorization", "Accept");
 
     /**
      * The parameter obtained {@link ServletConfig#getInitParameter(String)} to determine if legacy processing of
@@ -488,6 +491,11 @@ public abstract class HttpServlet extends GenericServlet {
 
         while (reqHeaderEnum.hasMoreElements()) {
             String headerName = reqHeaderEnum.nextElement();
+
+            if (isSensiveHeader(headerName)) {
+                continue;
+            }
+
             buffer.append(CRLF).append(headerName).append(": ").append(req.getHeader(headerName));
         }
 
@@ -499,6 +507,10 @@ public abstract class HttpServlet extends GenericServlet {
         resp.setContentLength(responseLength);
         ServletOutputStream out = resp.getOutputStream();
         out.print(buffer.toString());
+    }
+
+    private boolean isSensiveHeader(String headerName) {
+        return HEADER_SENSITIVE.contains(headerName);
     }
 
     /**
--- a/api/src/main/java/jakarta/servlet/http/HttpServlet.java
+++ b/api/src/main/java/jakarta/servlet/http/HttpServlet.java
@@ -492,7 +492,7 @@ public abstract class HttpServlet extends GenericServlet {
         while (reqHeaderEnum.hasMoreElements()) {
             String headerName = reqHeaderEnum.nextElement();
 
-            if (isSensiveHeader(headerName)) {
+            if (isSensitiveHeader(headerName)) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -501,6 +509,10 @@ protected void doTrace(HttpServletRequest req, HttpServletResponse resp) throws
         out.print(buffer.toString());
     }
 
+    private boolean isSensitiveHeader(String headerName) {
+        return HEADER_SENSITIVE.contains(headerName);
+    }
+
```
markt-asf, line 515: Does this really need a separate method?

markt-asf (changes requested):
The code formatter has been updated so I am afraid you'll need to reformat the code and update the PR.

It would be helpful if you included a changes entry for this in the spec document.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
dsandrade0: In this [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-trace) don't have the list of sensitive headers. Then include on this method because Java method will be seen by JIT to compiler separated of #doTrace and will be better performance.

markt-asf: All JIT is likely to do is in-line the method - which is what I am suggesting you do anyway. It isn't as if the separate method adds to the readability of the code.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -78,6 +80,7 @@ public abstract class HttpServlet extends GenericServlet {
 
     private static final String HEADER_IFMODSINCE = "If-Modified-Since";
     private static final String HEADER_LASTMOD = "Last-Modified";
+    private static final List<String> HEADER_SENSITIVE = Arrays.asList("Authorization", "Accept");
```
markt-asf, line 83: Why these two methods? `Authorization` I understand but not `Accept`. And why not `Cookie`?

api/src/main/java/jakarta/servlet/http/HttpServlet.java
gregw: Either in line it, or make it protected so that an extender can change what headers are considered safe.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
gregw: I agree that `Cookie` is a sensitive header, whilst no so much `Accept`.

I also think that any proxy headers (`X-Forwarded-*`, `Forwarded`,  `Proxy-Authorization`) should be protected so that trace can't be used to probe the internals of a data centre.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -488,6 +491,11 @@ protected void doTrace(HttpServletRequest req, HttpServletResponse resp) throws
 
         while (reqHeaderEnum.hasMoreElements()) {
             String headerName = reqHeaderEnum.nextElement();
+
+            if (isSensitiveHeader(headerName)) {
+                continue;
```
gregw, line 496: Should the headers be skipped, or should just there values be obscured?  RFC kind of says both.

gregw (changes requested):

api/src/main/java/jakarta/servlet/http/HttpServlet.java
dsandrade0: Maybe obscure can show that theses headers exists. Should the response knows theses headers are present?

api/src/main/java/jakarta/servlet/http/HttpServlet.java
dsandrade0: Then I will remove the class field HEADER_SENSITIVE and move all implementation to isSensitiveHeader method

api/src/main/java/jakarta/servlet/http/HttpServlet.java
gregw: I think leave it as skipping.  If somebody wants to see more then they can implement doTrace themselves.

markt-asf:
The review comments have been mostly addressed but I am going to apply this manually so I can make a few tweaks including:
- Adding the change to the change log section in the specification
- Re-wording the explanation of sensitive headers added to the spec
- Generate the default list of headers to exclude once and re-use it
- Make header comparison case insensitive
- Add Javadoc to `isSensitiveHeader()`
- Rename test methods
- Refactor test slightly to avoid NPE warning

markt-asf:
Greg's comments have been addressed so I am going to dismiss that review as I think not doing so will block the merge of the PR (even doing it manually). If I have missed anything, I'm happy to take an action to get it fixed.

markt-asf:
Tx. Merged via #519

PR #507, opened 2023-02-24
[Issue 133] Add HTTP PATCH

garydgregory:
https://github.com/jakartaee/servlet/issues/113


garydgregory:
My ECA shows a green checkmark and I've resubmitted, so I'm not sure what " The author(s) of the pull request is not covered by necessary legal agreements in order to proceed." is actually complaining about.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -394,6 +446,7 @@ protected void doOptions(HttpServletRequest req, HttpServletResponse resp) throw
         boolean ALLOW_GET = false;
         boolean ALLOW_HEAD = false;
         boolean ALLOW_POST = false;
+        boolean ALLOW_PATCH = false;
```
stuartwdouglas, line 449: You are missing the code to set this to 'true' if the method is present.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
```diff
@@ -543,6 +602,9 @@ protected void service(HttpServletRequest req, HttpServletResponse resp) throws
             maybeSetLastModified(resp, lastModified);
             doHead(req, resp);
 
+        } else if (method.equals(METHOD_PATCH)) {
```
stuartwdouglas, line 605: Can you add this at the end of the if statement? The more common methods should be at the top.

api/src/main/resources/jakarta/servlet/resources/web-app_2_4.xsd
```diff
@@ -614,6 +614,7 @@
     <xsd:simpleContent>
       <xsd:restriction base="j2ee:string">
 	<xsd:enumeration value="GET"/>
+	<xsd:enumeration value="PATCH"/>
```
stuartwdouglas, line 617: We can't just change a released XSD, we need a new version.

stuartwdouglas (dismissed):

stuartwdouglas:
With regards to the eca issue does the email used in the commits match the email used to sign the agreement?

api/src/main/java/jakarta/servlet/http/HttpServlet.java
garydgregory: Fixed.

api/src/main/java/jakarta/servlet/http/HttpServlet.java
garydgregory: Fixed

api/src/main/resources/jakarta/servlet/resources/web-app_2_4.xsd
garydgregory: Removed.

markt-asf:
Your work email was used for the commits. If you change that to your gmail address and then force push that should fix it.

garydgregory:
> Your work email was used for the commits. If you change that to your gmail address and then force push that should fix it.

Thank you @markt-asf ! Fixed and forced pushed.

I've now removed my +1 brain of farting...

stuartwdouglas (commented):
LGTM, although we probably also need a changelog entry.

garydgregory:
> I'll merge this shortly. It looks like the update to the formatter is going to impact one of the files changed here. I'll fix that in a separate PR and add the change log entry.

Thanks @markt-asf

markt-asf:
It is an API change so it will have to wait for Servlet 6.1.

It is available in the Tomcat 11 milestone releases - that includes a Servlet API JAR that includes `doPatch`. Usual caveats apply about using milestone releases.

garydgregory:
> It is an API change so it will have to wait for Servlet 6.1.
> 
> It is available in the Tomcat 11 milestone releases - that includes a Servlet API JAR that includes `doPatch`. Usual caveats apply about using milestone releases.

@markt-asf,

Thanks for the quick reply Mark.

Any thoughts on when to expect 6.1?

My goal is to push this through Jetty on which we have hard-coded dependencies.

markt-asf:
Best guess is late this year / early next.

PR #519, opened 2023-03-22
Fix #469 - Remove sensitive header on TRACE method

### LINKED ISSUE #469: AS per RFC 9110 (and earlier) HTTp TRACE response MUST not include sensitive headers

markt-asf:


markt-asf:
Force push to fix copyright in test case header

PR #533, opened 2023-06-13
Fix #431 - getParameter() may throw runtime exception

### LINKED ISSUE #431: multipart-config & getParameter clarification
I apologize if this issue has been brought up already, but I'm wondering if there are any plans to address this in a future spec. 

Scenario: There is a `multipart-config` servlet with a `max-file-size` specified, and a file, which is over the limit,  is uploaded to this servlet. The servlet then first calls `getParameter` when processing the request. 

During parsing of the parameters, the servlet tries to parse the multi-part form data for any parameters but encounters an issue with the file being too large. Some exception is thrown, and, since parsing was incomplete, `getParameter` returns null (even if it was for a valid parameter).

My question is how should the servlet implementation handle this scenario?

- Should null continue to be returned (current behavior)?
- Should an `IllegalStateException` or some other exception be thrown to indicate something went wrong? 
- Some other option (i.e `getParts` should be checked first as explained in ryanchapin's post below) ? 

For context, this was discovered through a JSF (MyFaces) upload application.  MyFaces checks postback requests by looking for a view-state parameter but the uploaded file is over the max-file-size.  Hence, `getParameter` is null.  JSF thinks this is a new request.  The getParameter is called by the implementation (so the application hasn't had a chance to call getPart/getParts yet)

If I'm missing something here or there's an existing workaround, please let me know. Otherwise, I'm hoping a future specification may address this scenario. 

I found a old blog post here, so it seems like this issue has exist for a while: https://www.ryanchapin.com/invocation-of-request-getparts-in-a-servlet-3-0-dopost-method-will-not-throw-illegalstateexception/

Thanks! 

References: 

1) https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletRequest.html#getParameter-java.lang.String-

2) https://javaee.github.io/javaee-spec/javadocs/javax/servlet/annotation/MultipartConfig.html 

3) 3.2 File upload  - Servlet 4.0 Spec

markt-asf:
Applies to all of the parameter methods.
Containers may provide options for alternative error handling.

I am leaving this in draft for others to provide feedback while I work on implementing this in Tomcat to check I haven't missed anything.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletRequest.java
+++ b/api/src/main/java/jakarta/servlet/ServletRequest.java
@@ -164,9 +164,22 @@ public interface ServletRequest {
      * If the parameter data was sent in the request body, such as occurs with an HTTP POST request, then reading the body
      * directly via {@link #getInputStream} or {@link #getReader} can interfere with the execution of this method.
      *
+     * <p>
+     * If not already parsed, calling this method will trigger the parsing of the query string, POSTed form data where the
+     * request body has content type is <code>application/x-www-form-urlencoded</code> and POSTed form data where the
+     * request body has content-type <code>multipart/form-data</code> and the Servlet is configured with a
+     * {@link jakarta.servlet.annotation.MultipartConfig} annotation or a <code>multipart-config</code> element in the
+     * deployment descriptor.
+     *
      * @param name a <code>String</code> specifying the name of the parameter
      *
      * @return a <code>String</code> representing the single value of the parameter
+     *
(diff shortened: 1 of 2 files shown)


stuartwdouglas (approved):

gregw (approved):

markt-asf:
Just an update to reassure folks I haven't forgotten about this. I'm working on the Tomcat implementation and keep getting side-tracked. I want to make sure the Tomcat implementation doesn't throw up any wrinkles that require tweaks to this PR before merging it.

PR #541, opened 2023-09-16
Add getAttribute method with generic return type

hakan-krgn:
This commit adds default getAttribute method with generic return type to ServletContext.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -527,6 +527,36 @@ public interface ServletContext {
      * @throws NullPointerException if the argument {@code name} is {@code null}
      *
      */
+    default <T> T getAttribute(String name, Class<T> clazz) {
+        return clazz.cast(this.getAttribute(name));
+    }
+
+    /**
+     * Returns the servlet container attribute with the given name, or <code>null</code> if there is no attribute by that
+     * name.
+     *
+     * <p>
+     * An attribute allows a servlet container to give the servlet additional information not already provided by this
+     * interface. See your server documentation for information about its attributes. A list of supported attributes can be
+     * retrieved using <code>getAttributeNames</code>.
+     *
+     * <p>
+     * The attribute is returned as a <code>java.lang.Object</code> or some subclass.
+     *
+     * <p>
+     * Attribute names should follow the same convention as package names. The Jakarta Servlet specification reserves names
+     * matching <code>jakarta.*</code>.
+     *
+     * @param name a <code>String</code> specifying the name of the attribute
+     *
+     * @return an <code>Object</code> containing the value of the attribute, or <code>null</code> if no attribute exists
+     * matching the given name.
+     *
+     * @see ServletContext#getAttributeNames
+     *
+     * @throws NullPointerException if the argument {@code name} is {@code null}
+     *
+     */
     public Object getAttribute(String name);
 
     /**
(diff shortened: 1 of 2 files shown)


stuartwdouglas:
If we are going to do this shouldn't we also do the same for `ServletRequest`? It would also need a changelog entry.

arjantijms:
Is passing in the class really needed here?

The following should work as well, should it not?

```java
<T> T getAttribute(String name);
```

Otherwise the call sites would have

```java
Foo foo = (Foo) getAttribute(name);
```

vs

```java
Foo foo = getAttribute(name, Foo.class);
```

Which doesn't seem to be much of an improvement tbh.

stuartwdouglas:
We already have `public Object getAttribute(String name)`, which has the same erasure as `<T> T getAttribute(String name)`.

We can't really modify the existing method without breaking compatibility.

arjantijms:
> We can't really modify the existing method without breaking compatibility.

Just wondering, in what situation would it break?

hakan-krgn:
> Is passing in the class really needed here?
> 
> The following should work as well, should it not?
> 
> ```java
> <T> T getAttribute(String name);
> ```
> 
> Otherwise the call sites would have
> 
> ```java
> Foo foo = (Foo) getAttribute(name);
> ```
> 
> vs
> 
> ```java
> Foo foo = getAttribute(name, Foo.class);
> ```
> 
> Which doesn't seem to be much of an improvement tbh.

Yes, that's right. I was going to change return type of ``public Object getAttribute(String name)`` method with generic type but I didn't want to change the project structure, that's why I added a default method with generic type.




> We already have `public Object getAttribute(String name)`, which has the same erasure as `<T> T getAttribute(String name)`.
> 
> We can't really modify the existing method without breaking compatibility.

What kind of compatibility issue are we talking about?

stuartwdouglas:
So here is an example:

```

   public static Object getAttribute(String name){
       return null;
   }

//    public static <T> T getAttribute(String name) {
//        return null;
//    }

    static void a(Object a) {
        System.out.println("Object");
    }

    static void a(String a) {
        System.out.println("String");
    }

    static void a(Integer a) {
        System.out.println("int");
    }

    public static void main(String... args) {
        a(getAttribute("ignored"));
    }
```

If you run the code as it is then you will get the output 'Object'. If you comment out the original version and uncomment the generic version then you will get a compile error because it can't pick between the int and string versions. If you then comment out the Integer version of a() you will get the output 'String'.

arjantijms:
Totally offtopic but maybe a default for such cases in the JLS would have helped:

```java
    public static <T> T default Object getAttribute(String name) {
        return null;
    }
```

markt-asf:
> ```java
> Foo foo = (Foo) getAttribute(name);
> ```
> 
> vs
> 
> ```java
> Foo foo = getAttribute(name, Foo.class);
> ```
> 
> Which doesn't seem to be much of an improvement tbh.

Give the above, is this a change we want to pursue? Do we want to look at an alternative such as:

```java
<T> T getTypedAttribute(String name);
``` 

or do we just leave the API as-is for now?

stuartwdouglas:
I am leaning towards leave it as-is, although I am not really sure how big an issue the compatibility problem would be, my example was somewhat contrived and it would probably be very rare for it to actually be a problem.

markt-asf:
Closing on the basis we will leave this as-is for now. We can always revisit his later if there is interest.

PR #543, opened 2023-09-26
Fix #538 - Make HTTP/2 push support optional

### LINKED ISSUE #538: Drop support for server push
Given that [Chrome has dropped support for server push](https://developer.chrome.com/blog/removing-push/) I'd like to propose the following:

- deprecate `HttpServletRequest.newPushBuilder()`
- deprecate `HttpServletRequestWrapper.newPushBuilder()`
- deprecate `PushBuilder`
- document that `newPushBuilder()` always returns null

Remove the deprecated methods in Servlet 7.0

markt-asf:


api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
```diff
@@ -288,11 +288,15 @@ public String toString() {
...
      *
-     * @return a {@link PushBuilder} for issuing server push responses from the current request, or null if push is not
-     * supported
+     * @return a {@link PushBuilder} for issuing server push responses from the current request, or {@code null} if push is
+     * not supported. Note that some implementations may opt not to support server push and will therefore always return
+     * {@code null}
      *
      * @since Servlet 4.0
+     *
+     * @deprecated Use early hints via TBD instead
```
gregw, line 297: Not a huge fan of the TBD.   
```suggestion
     * @deprecated In favor of 103 early hints
```

When we have an early hint mechanism, then we can remember to update this just as well as remembering to remove the TBD

gregw (approved):
Looks good... but not a fan of the TBD

api/src/main/java/jakarta/servlet/http/HttpServletRequest.java
markt-asf: Fair enough. I'll update the PR.

PR #547, opened 2023-10-02
Fix #546 - mechanism to access HttpSession outside of request

### LINKED ISSUE #546: Enable HttpSession to be used outside of the scope of an HttpServletRequest
This is primarily intended to provide a mechanism to address https://github.com/jakartaee/websocket/issues/175 but the solution is intended to be a general one.

There has been some prior discussion of this issue on the Servlet dev list:
https://www.eclipse.org/lists/servlet-dev/msg00566.html

This issue is to track the implementation of the solution that was broadly agreed on that thread.

I'll provide a PR along the lines of the solution discussed above. I expect there will be a few iterations to refine that solution before it is merged.

markt-asf:


joakime:
This seems like a partial solution.

The `HttpSession` accessed from this new accessor can do what?
Is it read-only? (I hope so, otherwise syncing changes into/out of the HttpSession when held long term in a place like websocket would be a pain in the ass)

markt-asf:
It provides full read/write access to the session.

The use of the Consumer interface (see the dev list thread for the details) is so WebSocket isn't holding a reference to the actual session.

api/src/main/java/jakarta/servlet/http/HttpSession.java
```diff
@@ -213,4 +219,24 @@ public interface HttpSession {
...
     public boolean isNew();
 
+    /**
+     * Provides a mechanism for applications to interact with the {@code HttpSession} outside of the scope of an HTTP
+     * request.
+     * <p>
+     * To use this mechanism, applications call the {@code Consumer#accept(Object)} method on the {@code Consumer} instance
+     * returned from this method and pass an application provided instance of {@code Consumer<Session>}. During that call,
+     * the container will call the the {@code Consumer#accept(Session)} method of the application provided
+     * {@code Consumer<Session>} passing the {@code HttpSession} object.
```
gregw, line 229: I think we need to make it clear that the `HttpSession` passed to the consumer may be a new instance and that the session may have been passivate/activated since the accessor was created:

```suggestion
     * To use this mechanism, applications call the {@code getAccessor()} method on {@code HttpSession} instance that will be accessed.
     * This call must be made from within the scope of an HTTP request and will return a  {@code Consumer<Consumer<HttpSession>>}
     * that then can be used outside of the scope of a HTTP request to access the session.   Such access is obtained by passing a 
     * {@code Consumer<HttpSession>>} to the {@code Consumer#accept(Object)} method on the {@code Consumer} instance
     * returned from this method.  During that call, the container will call the the {@code Consumer#accept(Session)} method of the 
     * application provided {@code Consumer<Session>} passing an {@code HttpSession} object that represents the same 
     * {@code HttpSession} that the accessor was obtained from, but it may be a different instance and have been passivated then activated
     * since the accessor was obtained.  If the session has been invalidated before the call, then the consumer will be called with an invalid 
     * {@code HttpSession} instance representing the original {@code HttpSession}. 
     
```

gregw (changes requested):

gregw:
One way to think of this is that the inner consumer takes the place of a Servlet accessing a session. The outer consumer takes the place of a season manager that scopes the call to the "Servlet" by ensuring the season is activated and updating access time/count etc and then counts decremented after the call

api/src/main/java/jakarta/servlet/http/HttpSession.java
```diff
@@ -213,4 +219,24 @@ public interface HttpSession {
...
+     * <p>
+     * For the purposes of session access, validity, passivation, activation etc. the container behaves as if the call the
+     * container makes to the {@code Consumer#accept(Session)} method of the application provided {@code Consumer<Session>}
+     * instance occurs during the processing of an HTTP request.
+     *
+     * @return A container provided {@code Consumer} instance that accepts application provided {@code Consumer<Session>}
+     * instances to enable the application to interact with the HTTP session outside of the scope of an HTTP request or
+     * {@code null} if access to the session is not supported outside of an HTTP request
+     */
+    default public Consumer<Consumer<HttpSession>> getAccessor() {
```
stuartwdouglas, line 239: I think we also need to allow this method to fail somehow, as some cluster based solutions might not want to allow access outside the scope of a HTTP request.

gregw: Isn't that the null return?  Or do you want the returned consumer be able to fail at a later time?

stuartwdouglas: Sorry, I missed that, too early in the morning and had not had my coffee yet :-)

joakime:
Has anyone bounced this idea off the CDI folks?
Feels like they will have to change their servlet integration implementations to support this.

api/src/main/java/jakarta/servlet/http/HttpSession.java
```diff
@@ -213,4 +219,30 @@ public interface HttpSession {
...
+     * To use this mechanism, applications call the {@code getAccessor()} method on {@code HttpSession} instance that will be accessed.
+     * This call must be made from within the scope of an HTTP request and will return a  {@code Consumer<Consumer<HttpSession>>}
+     * that then can be used outside of the scope of a HTTP request to access the session.   Such access is obtained by passing a 
+     * {@code Consumer<HttpSession>>} to the {@code Consumer#accept(Object)} method on the {@code Consumer} instance
+     * returned from this method.  During that call, the container will call the the {@code Consumer#accept(Session)} method of the 
+     * application provided {@code Consumer<Session>} passing an {@code HttpSession} object that represents the same 
+     * {@code HttpSession} that the accessor was obtained from, but it may be a different instance and have been passivated then activated
+     * since the accessor was obtained.  If the session has been invalidated before the call, then the consumer will be called with an invalid 
+     * {@code HttpSession} instance representing the original {@code HttpSession}. 
+
```
gregw, line 235: Blank line.... so probably should fill it with:
```suggestion
     * {@code HttpSession} instances passed to the provided {@code Consumer<Session>} must not be used or referenced outside of the scope of the 
     * call that provided them.
```

gregw (approved):
Generally approve, although happy for the text to be generally tweaked and/or formatted.

gregw (changes requested):
Chatting to @janbartel about this and she pointed out that it would be better to act like other `HttpSession` methods and throw ISE if the session is invalid.  This is both for calling `getAccessor()` and `accept(Consumer<HttpSession>)`.     So we should change the text about passing an invalid session.

We also realized that there are issues with sessions that change their ID.  If you get an accessor to a session, which is then locally passivated, then it changes ID on another node in the cluster, then you try to use the accessor, it will not find the session.     This is probably a good thing, as sessions change their IDs over authentication boundaries and we don't want code that knew of an unauthenticated session having access to the authenticated one.  So verbage might be needed for this.

Finally,  it might be clearer to drop one of the `Consumer`s and instead have:

```java
interface HttpSession {
    // ...
    interface Accessor {
        access(Consumer<HttpSession>) throws IllegalStateException;
    }
    Accessor getAccessor();
}
```

api/src/main/java/jakarta/servlet/http/HttpSession.java
```diff
@@ -213,4 +219,30 @@ public interface HttpSession {
...
+     * container makes to the {@code Consumer#accept(Session)} method of the application provided {@code Consumer<Session>}
+     * instance occurs during the processing of an HTTP request.
+     *
+     * @return A container provided {@code Consumer} instance that accepts application provided {@code Consumer<Session>}
+     * instances to enable the application to interact with the HTTP session outside of the scope of an HTTP request or
+     * {@code null} if access to the session is not supported outside of an HTTP request
+     */
+    default public Consumer<Consumer<HttpSession>> getAccessor() {
+        return null;
+    }
```
gregw, line 247:
```suggestion
    /**
     * An accessor for applications to interact with a {@code HttpSession} outside of the scope of an HTTP request.
     */
    interface Accessor {
        /**
         * Call to access the {@code HttpSession} used to obtain this {@code Accessor} from outside the scope of a HTTP request.
         * <p>
         * When called, the container will call the {@link Consumer#accept(Object)} method of the {@code sessionConsumer}
         * passed by the application, with an {@code HttpSession} object that represents the same {@code HttpSession} that this
         * {@code Accessor} was obtained from.
         * <p>
         * The passed {@code HttpSession} may be the same instance as used to obtain this {@code Accessor}, or a new instance 
         * for the session which has possibly been passivated and activated since the {@code Accessor} was obtained.
         * <p>
         * The passed {@code HttpSession} may be shared concurrently with other {@code Accessor}s for the same session,
         * other calls to this {@code Accessor} and/or {@link jakarta.servlet.Servlet}s and {@link jakarta.servlet.Filter}s.
         * <p>
         * The passed {@code HttpSession} must not be used or referenced outside the scope of the call
         * to the {@link Consumer#accept(Object)} method of {@code sessionConsumer}.
         * <p>
         * For the purposes of session access, validity, passivation, activation etc. the container behaves as if the call 
         * to the {@link Consumer#accept(Object)} method of {@code sessionConsumer} occurs during the processing of an HTTP 
         * request for the same session.
         *
         * @param sessionConsumer the application provided {@link Consumer} of {@code HttpSession} to access the session.
         * @throws IllegalStateException if this method is called on an invalidated session or if the session ID has
         *                               changed since the {@code Accessor} was obtained.
         */
         void access(Consumer<HttpSession> sessionConsumer);
    }

    /**
     * Provides a mechanism for applications to interact with this session outside the scope of an HTTP request.
     *
     * @return A container provided {@link Accessor} for this session or {@code null} if access is not supported outside 
     *         an HTTP request.
     * @throws IllegalStateException if this method is called on an invalidated session
     */
    default Accessor getAccessor() {
        return null;
    }

```

gregw (changes requested):

markt-asf:
Why should changing the session ID always prevent the Accessor from continuing to work? I can see some circumstances where that could happen but there are some cases where the Accessor could continue to work.

I also don't see the security concern if an application uses the same session object (with different IDs) before and after authentication. Any nefarious client activity (such as session fixation attacks) can only reference the session by ID and the change in ID is sufficient to protect against that sort of behaviour.

gregw:
@markt-asf you're right that it could sometimes be implemented to persist over an ID change, but it cannot be done so generally.   Specifically in a cluster if the session is locally passivated, and the ID is changed on another node, then the semantics are really poorly defined.

Thus I was going for well defined semantics that are easy to implement.  The accessor keeps the session ID locally and if it can find a valid session when it is accessed, then all is good.

If we allow an accessor to remain valid over an ID change, then we either are going to have inconsistent behaviour depending on the actual session implementation or deployment, OR we will need to implement some complex mechanism to update all accessors in a cluster.

I do think it is of slight concern that a websocket connection established before a session is authenticated may be able to maintain access to the session after authentication.   I don't see any huge vulnerability immediately, but the whole point of changing a session's ID at authentication is that the session then becomes accessible only to clients that are given that ID after authentication.   Put another way, when code calls a method to change the session id, it is doing so to reduce the number of references to that session to zero, and it can then hand out new references as it sees fit.    Having accessors bypass that mechanism feels like a violation of an important contract.    Note that this is not about what happens if a `Consumer<HttpSession>` is actively being called when a session ID changes, as that is equivalent to what happens if there are concurrent request.    Calling the accessor is semantically equivalent to a new request arriving and unless the client of that new request has explicitly been given the new session ID, then it cannot access the session.

janbartel:
FWIW I've implemented the `HttpSession.Accessor` interface in an experimental jetty 12 branch and it was easy to do. I'll add some more test cases to be sure, but so far it looks very promising.

markt-asf:
Implementing this in Tomcat was also simple.
PR LGTM. I think we can merge this. @gregw if you can approve the changes you requested that should enable this to me merged with having to dismiss your review.

gregw (approved):

markt-asf:
Forced push was just to rebase to fix merge conflicts in spec doc

PR #549, opened 2023-10-08
Add request attribute for secure_protocol.

### LINKED ISSUE #130: Please add servlet request attribute for SSL/TLS protocol
Since the early days of the servlet specification, the following request attributes have been provided to applications wishing to check the details of the remote user's SSL/TLS connection:

javax.servlet.request.cipher_suite
javax.servlet.request.key_size
javax.servlet.request.ssl_session_id
javax.servlet.request.X509Certificate

It would also be nice to be able to see the SSL/TLS protocol (e.g. SSLv3, TLSv1, TLSv1.2, etc.).

I propose the following request attribute to be added to the specification and required by compliant containers:

javax.servlet.request.ssl_protocol (consistent with existing .ssl_* attribute names)
or
javax.servlet.request.tls_protocol (consistent with likely future protocols)
or
javax.servlet.request.secure_protocol (protocol-name agnostic)

The list of acceptable values should match the JVM's naming scheme for SSL/TLS protocol, but not be restricted to the protocols supported by the underlying Java Runtime Environment (in the event that a crypto provider other than JSSE – e.g. OpenSSL, etc. – is being used to provide the secure connection instead of the JRE).

The level-of-effort to modify the specification should be minimal, and the burden on container implementers should be similarly minimal, since the existing request attributes must already be populated in the request attributes, and the SSL/TLS protocol is readily available from the same source as the aforementioned attribute values.

ChristopherSchultz:
This fixes #130

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -1729,6 +1729,10 @@ must expose the following attributes to the servlet programmer:
 |Attribute Name
 |Java Type
 
+|protocol e.g. TLSv1.3
+|`jakarta.servlet.request.secure_protocol`
+|`String`
+
 |cipher suite
 |`jakarta.servlet.request.cipher_suite`
 |`String`


markt-asf:
Note a couple of alternative attribute names were suggested in #130. I'm happy with this particular name but could live with either of the alternatives.
If there are no objections, I intend to merge this in a week or so for inclusion in 6.1.

ChristopherSchultz:
I ultimately chose `secure_protocol` for several reasons:

1. `ssl_protocol` is just wrong, since SSL is no longer used
2. `tls_protocol` may be wrong in the future, once we switch to TBD
3. `tls_protocol` may also be wrong for some other protocol, say, sftp or whatever
4. Neither `cipher_suite` nor `key_size` have any `ssl_` prefix as `ssl_session_id` does
5. Using `protocol` without a prefix would be confusing with respect to HTTP/1.1, etc.; this is intended to be the *security protocol* in use for the request

markt-asf:
Tx for the PR. I'll follow-up shortly with a PR to add an entry to the change log etc.

PR #559, opened 2023-11-04
Fix wrong description in Chapter4.4

OnlyWick:


markt-asf:
This change is not correct. The "instance of" is important here. Also grammatically, "an instance" would need to change to "a servlet" or "a filter".

What problem do you see here that you are trying to solve?

OnlyWick:
These methods accept a parameter of Class type, `Class <? extends Servlet> servletClass`.

OnlyWick:
It is usually used like this, `ServletContext.addServlet("xxx", XXX.class)`. Is it not correct?

PR #560, opened 2023-11-04
Fix backtick error in Chapter4.4

OnlyWick:


markt-asf:
Thanks for catching these.

PR #565, opened 2023-11-16
Fix #563 - version attributes not replaced in web.xml examples

### LINKED ISSUE #563: Descriptor example in Servlet 6.0 doc is bad
The example descriptor in the published Servlet 6.0 documentation is incorrect.

See document at https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#a-basic-example

It appears like the following ...

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee
         web-app_{spec-version-underscore}.xsd"
         version="{spec-version}">
```

The `{spec-version-underscore}` and `{spec-version}` were apparently not expanded properly.

markt-asf:


pzygielo:
- Created #864 which I think this one caused.

PR #572, opened 2024-01-09
Fix #571 - clarify behaviour of Cookie.setAttribute(String, String)

### LINKED ISSUE #571: Clarify Cookie attribute behavior for empty and null values
#175 introduced the support for Cookie attributes in order to generally address the need for attributes such as "SameSite".

Right now, [third party cookie restrictions are being rolled out in Google Chrome](https://developers.google.com/privacy-sandbox/3pcd) and many application will need to add the "Partitioned" attribute to keep things working. I have tried to use the attributes API for this purpose and failed:

```java
Cookie cookie = new Cookie("test", "value");
// set attribute value
cookie.setAttribute("attribute", "attrvalue");
// set empty attribute
cookie.setAttribute("Partitioned", "");
// set empty attribute and then null it out
cookie.setAttribute("nullable", "value");
cookie.setAttribute("nullable", null);
response.addCookie(cookie);
```

| Container             | Cookie header                                               |
|-----------------------|-------------------------------------------------------------|
| Tomcat 10.1.17        | `Set-Cookie: test=value; attribute=attrvalue; Partitioned=` |
| Jetty 12.0.5          | `Set-Cookie: test=value; attribute=attrvalue`               |
| Undertow 2.3.10.Final | `Set-Cookie: test=value`                                    |

Note, I have removed the "nullable" attribute in my tests with Tomcat 10.1.17 since it currently fails for `null` attribute values with the following:

```
java.lang.NullPointerException: Cannot invoke "String.toCharArray()" because "value" is null
	at org.apache.tomcat.util.http.Rfc6265CookieProcessor.validateAttribute(Rfc6265CookieProcessor.java:270) ~[tomcat-embed-core-10.1.17.jar:10.1.17]
	at org.apache.tomcat.util.http.Rfc6265CookieProcessor.generateHeader(Rfc6265CookieProcessor.java:193) ~[tomcat-embed-core-10.1.17.jar:10.1.17]
```

Could you clarify the cookie attributes behavior for the following cases:

1. Add a cookie attribute with a `null` value - should this remove the attribute from the map or should it be rejected?
2. Add a cookie attribute with an empty value - ideally, we should support the `Partitioned` case (so no additional `=` sign)

markt-asf:


api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -441,9 +441,19 @@ public boolean isHttpOnly() {
      * except for <code>version</code>. E.g. when <code>cookie.setAttribute("domain", domain)</code> is invoked, then
      * <code>cookie.getDomain()</code> should return exactly that value, and vice versa.
      *
+     * <p>
+     * Attributes with a value of the empty string will appear as {@code "attribute-name="} in any {@code Set-Cookie} header
```
joakime, line 445:
```suggestion
     * Attributes with a value of the empty string will appear as {@code "attribute-name"} (a cookie without a value or equals sign) in any {@code Set-Cookie} header
```

api/src/main/java/jakarta/servlet/http/Cookie.java
```diff
@@ -441,9 +441,19 @@ public boolean isHttpOnly() {
...
+     * generated from the {@code Cookie} instance unless special handling is required as defined in the following paragraph.
+     *
+     * <p>
+     * Attributes that are defined as not having a value, such as {@code HttpOnly}, must set with a value such that {@code
+     * Boolean.parseBoolean(attribute-value)} evaluates to {@code true} for the attribute to be included in any
+     * {@code Set-Cookie} header generated from the {@code Cookie} instance, otherwise the attribute is not included.
```
joakime, line 451: This whole block can be dropped if the verbiage above is done instead.

This makes the results of `setAttribute("name", "")` produce parsing rules that comply today with things like `Partitioned`, and any future attributes without values too.

Perhaps this can even provide a code block example.

``` java
<code>
setAttribute("name", "value"); // becomes "name=value"
setAttribute("name", ""); // becomes "name"
setAttribute("name", null); // removes "name" from the cookie
setAttribute("HttpOnly", ""); // becomes "HttpOnly"
setAttribute("Partitioned", ""); // becomes "Partitioned"
setAttribute("SameSite", "Strict"); // becomes "SameSite=Strict"
</code>
```

Pointing out what happens?

joakime (changes requested):
Simplify the verbiage a bit, make it more generic, yet still spec compliant.

api/src/main/java/jakarta/servlet/http/Cookie.java
markt-asf: Applied with the minor change of "...cookie **attribute** without a value..."

joakime (approved):

volosied:
@markt-asf  Please include this in 6.1, if possible. Thank you

PR #575, opened 2024-02-27
TCK fixes

markt-asf:
This PR fixes the following issues I found while trying to build the latest JSP TCK (that depends on the Servlet TCK).

The Servlet API version the TCK depended on needed to be updated to the latest 6.1.0.

One of the tests needed some casts added to resolve the ambiguity between `setCharacterEncoding(String)` and `setCharacterEncoding(Charset)`.

I think these changes are uncontroversial but since it is the TCK I intend to allow at least 24 hours for review before I merge these changes.

### DIFF
--- a/tck/tck-runtime/pom.xml
+++ b/tck/tck-runtime/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 1997, 2022 Oracle and/or its affiliates and others.
+    Copyright (c) 1997, 2024 Oracle and/or its affiliates and others.
     All rights reserved.
 
     This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


olamy:
LGTM.
But it would be good to have tck artifacts deployed as well.
The tag is here, the pom versions have been changed as well but no artifacts available.
I'm currently  working on removing everything servlet tck related from the main/platform TCK but without the artifacts available it;s not possible (well I can't ask people to rebuild from sources before running TCK :))

markt-asf:
I'll see about getting an M1 deployed.

markt-asf:
The API and the TCK are likely to have divergent point releases. For example if we need to exclude a test from the TCK as the result of a challenge. I should probably revert that change in the POM so they remain independent. I'll update the PR.

markt-asf:
M1 TCK release is in staging repo. Tagged as 6.1.0-RELEASE-tck. CI system is setup to do snapshot and staging builds. I'm seeing quote a few failures but that could well be configuration issues on my side. I'll open new issues as required as I look into those failures.

Issue #578, opened 2024-03-02
jakarta.servlet-api.jar MANIFEST.MF contains path to builder's current directory


hboutemy:
in `META-INF/MANIFEST.MF`, for entries `META-INF/LICENSE.md` and `META-INF/NOTICE.md`, there is a syntax that says `=/home/jenkins/agent/workspace/1_servlet-api-build-and-stage/` + file name

extract from `jakarta.servlet-api-6.1.0-M2.jar` `META-INF/MANIFEST.MF`:
```
3_1.xsd,META-INF/LICENSE.md=/home/jenkins/agent/workspace/1_servlet-api
-build-and-stage/LICENSE.md,META-INF/NOTICE.md=/home/jenkins/agent/work
- space/1_servlet-api-build-and-stage/NOTICE.md
```

see full reference https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/jakarta/servlet/jakarta.servlet-api/README.md

olamy:
@hboutemy I'm not clear with your report. you are mentioning `META-INF/LICENSE.md` and `META-INF/NOTICE.md` but I cannot see any issues with those files,
But I can see issues with `META-INF/MANIFEST.MF`
```
unzip -p jakarta.servlet-api-6.1.0-M2.jar META-INF/MANIFEST.MF | grep jenkins
 3_1.xsd,META-INF/LICENSE.md=/home/jenkins/agent/workspace/1_servlet-api
 -build-and-stage/LICENSE.md,META-INF/NOTICE.md=/home/jenkins/agent/work
```

Is it your concern?

hboutemy:
@olamy yes, I was not clear, updated the description to clarify
any idea what creates this?

olamy:
@hboutemy the guilty is osgi instruction `Include-Resource`. 
And I'm not quite sure how to fix it actually. 
The easiest  fix is to change the bundle plugin instruction to `<_removeheaders>Bnd-LastModified,Include-Resource</_removeheaders>`
@markt-asf do we need the `Include-Resource` osgi instruction to be there?
Because ATM it's really wrong and the whole content including 
```
Include-Resource: jakarta/servlet/LocalStrings.properties=src/main/java/
 jakarta/servlet/LocalStrings.properties,jakarta/servlet/LocalStrings_fr
 .properties=src/main/java/jakarta/servlet/LocalStrings_fr.properties
```
[...]

PR #579, opened 2024-03-12
Issue #578 make produced META-INF/MANIFEST.MF reproducible

olamy:


markt-asf:
Having researched the header a little, I think we can just drop it. It isn't an OSGi standard header, it is BND specific and serves no useful purpose I can see.

olamy:
oops my bad. I didn't check correctly.
Sorry for that.
I have removed completely the OSGI header.

olamy:
Nothing controversial here. So merging it.

PR #585, opened 2024-03-18
Align HEAD TCK test with RFC 9110, section 9.3.2

markt-asf:
Ignore content-length and vary headers

This should be fairly uncontroversial but it is a change so I am intending to allow at least 24 hours, and hopefully longer, for review.

### DIFF
--- a/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
+++ b/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2021 Oracle and/or its affiliates and others.
+ * Copyright (c) 2007, 2024 Oracle and/or its affiliates and others.
  * All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -64,9 +64,9 @@ public class HttpServletRequestTests extends HttpRequestClient {
     // ------------------------------
     /*
    * @testName: getAttributeNamesTest
-   * 
+   *
    * @assertion_ids: Servlet:JAVADOC:573
-   * 
+   *
    * @test_Strategy: Servlet verifies attributes
    *
    */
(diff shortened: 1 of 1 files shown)


tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
```diff
@@ -685,23 +685,16 @@ public void doHeadTest() throws Exception {
...
-//            }
 
             List<Header> headersGet = responseGet.getResponseHeaders();
             Set<String> headersToMatch = headersGet.stream()
                     .map(Header::getName)
+                    // Filter date header as it will change between requests
                     .filter(name -> !name.toLowerCase(Locale.ENGLISH).equals("date"))
+                    // Filter headers for which a value is determined during content generation
+                    // RFC 9110, section 9.32
+                    .filter(name -> !name.toLowerCase(Locale.ENGLISH).equals("content-length"))
```
olamy, line 696: you will need to ignore `content-length` later as well or this will fail as 

```
diff --git a/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java b/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
index 9ca4590..3db5fdc 100644
--- a/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
+++ b/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
@@ -699,8 +699,8 @@ public class HttpServletRequestTests extends HttpRequestClient {
 
             List<Header> headersHead = responseHead.getResponseHeaders();
             for (Header header : headersHead) {
-                if (header.getName().equalsIgnoreCase("date")) {
-                    // Skip date header
+                if (header.getName().equalsIgnoreCase("date") || header.getName().equalsIgnoreCase("content-length")) {
+                    // Skip date and content-length header
                     continue;
                 }
                 if (!headersToMatch.remove(header.getName())) {

```

olamy (changes requested):

tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletrequest/HttpServletRequestTests.java
markt-asf: Thanks for spotting that. I've pushed an updated fix.

olamy (approved):

Issue #603, opened 2024-03-27
Need to update schema for 6.1.0 release


markt-asf:

markt-asf:
Fixed by #604

PR #604, opened 2024-03-27
Add Jakarta EE 11 schema

markt-asf:
These should be the final versions but they weren't all available in the jakarta-schemas repository. I have provided PRs to jakarta-schemas and built the schemas required for the Servlet AOI locally. There is a (slim) chance that we'll need to update these again before the final release.

This should be uncontroversial as the 6.1 schemas are just the 6.0 schemas with a version update and one minor comment clarification. Given that and the looming deadline, I intend to merge immediately.


markt-asf:
This fixes #603

PR #620, opened 2024-05-09
Update groupId

markt-asf:


tck/tck-runtime/pom.xml
```diff
@@ -89,7 +89,7 @@
             <artifactId>arquillian-junit-core</artifactId>
         </dependency>
         <dependency>
-            <groupId>jakarta.servlet</groupId>
+            <groupId>jakarta.tck</groupId>
```
olamy, line 92: Can't we use jakarta.tck.servlet for all related servlet tck code?

markt-asf: I'm trying to remain consistent with the other TCKs. The groupId is one of the few things that does appear to be consistent (at the moment). The TCKs are still a bit of a moving target and I'm not getting any answers to my questions about what the Maven co-ordinates, package names etc should be. I don't have a strong opinion on what they should be but I do have a strong opinion that they should be consistent. That consistency has already been a huge help as I was able to essentially copy and paste the CI job to stage the TCK to the download directory from the authorization project.

olamy: fair enough. artifactId are `servlet-...` so that's good enough. no worries

PR #621, opened 2024-05-12
fix the build as 6.1.0 is not available in Maven central

olamy:

### DIFF
--- a/tck/tck-docs/userguide/pom.xml
+++ b/tck/tck-docs/userguide/pom.xml
@@ -28,7 +28,7 @@
     <groupId>org.glassfish</groupId>
     <artifactId>tck_servlet</artifactId>
     <packaging>pom</packaging>
-    <version>6.1.0</version>
+    <version>6.1.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


markt-asf:
Please revert. 6.1.0 is available from the staging repository.

olamy:
Sure but please provide the url of staging repo to help people building the project (we should have it in the pom if it's mandatory to build the project)
Is this a final release or a staged release? I would like to avoid people downloading a non snapshot ja which could potentially change.

markt-asf:
Good point. The TCK JAR could well change before the final release (hopefully not but you never know) and I agree downloading a non-snapshot JAR that is might change is not good. I suspect this will require a little more hoop jumping for the CI system so we don't end up with a released artifact that depends on a SNAPSHOT but let's see if we can address that in the CI system. I withdraw my request to revert this.

arjantijms:
> Sure but please provide the url of staging repo to help people building the project (we should have it in the pom if it's mandatory to build the project) 

It is in the pom already, and you can activate it via the "-Pstaging" parameter.

See the parent pom;

```xml
    <parent>
        <groupId>org.eclipse.ee4j</groupId>
        <artifactId>project</artifactId>
        <version>1.0.9</version>
        <relativePath/>
    </parent>
```

PR #654, opened 2024-06-11
Merge 6.1.0 release into master

arjantijms:
According to https://github.com/jakartaee/servlet/issues/651

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -30,7 +30,7 @@
 
     <groupId>jakarta.servlet</groupId>
     <artifactId>jakarta.servlet-api</artifactId>
-    <version>6.1.0-SNAPSHOT</version>
+    <version>6.1.0</version>
 
     <name>Jakarta Servlet</name>
     <url>https://projects.eclipse.org/projects/ee4j.servlet</url>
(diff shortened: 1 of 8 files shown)


pzygielo:
Could the error https://github.com/jakartaee/servlet/actions/runs/9227838949/job/25390577146?pr=654#step:5:2728 be fixed within this PR to not pollute master with non-buildable state?

markt-asf:
Not sure we really need to do this. I think renaming the 6.1.0 branch to 6.1.x and updating master to 6.2.0(-SNAPSHOT) would get us to where we need to be.

arjantijms:
> Not sure we really need to do this. 

It's partially common practice, but also the exact commit that made 6.1.0 a fact would not be in the 6.2.0 branch then, right? Not sure if that's an issue by itself.

But most (all?) projects do it like this.

markt-asf:
I guess I am -0 on this. I don;t really see the point but no objection. I'll rename the 6.1.0 branch to 6.0.x once you are done.

arjantijms:
@markt-asf it is done

markt-asf:
Tx. Branch renamed. I'll update the master branch to 6.2.x shortly.

Issue #660, opened 2024-06-20
TCK for Servlet 6.1 invalid error code  in servlet.tck.api.jakarta_servlet_http.httpservletresponse


janbartel:
https://github.com/jakartaee/servlet/blob/master/tck/tck-runtime/src/main/java/servlet/tck/api/jakarta_servlet_http/httpservletresponse/servlet_jsh_httpservletresponse_web.xml#L74 maps error code `411` to `servlet.tck.api.jakarta_servlet_http.httpservletresponse.ServletErrorPage`, however the server side part of the test at https://github.com/jakartaee/servlet/blob/master/tck/tck-runtime/src/main/java/servlet/tck/common/response/ResponseTests.java#L1079 is doing a `sendError` call for `HttpServletResponse.SC_GONE`, which is equivalent to `410`, so the `ServletErrorPage` will never be invoked.

The same error also affects:

- https://github.com/jakartaee/servlet/blob/master/tck/tck-runtime/src/main/java/servlet/tck/pluggability/api/jakarta_servlet_http/httpservletresponse/servlet_pluh_httpservletresponse_web-fragment.xml#L77
- https://github.com/jakartaee/servlet/blob/master/tck/tck-runtime/src/main/java/servlet/tck/pluggability/api/jakarta_servlet_http/httpservletresponsewrapper/HttpServletResponseWrapperTests.java

markt-asf:
I do not believe that it is intended that this test uses `ServletErrorPage`.

`ServletErrorPage` returns a plain text response but this test is expecting a content type of `text\html`.

The test expects the error page to echo the message passed to `sendError(int sc, String msg)` but `ServletErrorPage` only echos this message if it contains the text `error page invoked` which it does not in this case.

The test is only relying on documented behaviour of `HttpServletResponse.sendError(int, String)` (use the provided status code, send on HTML response, send the message in the response) and does not require a custom error page for this.

The use of status code 411 appears to be solely for the `sendError_StringErrorPageTest`.

I am curious why you think this test should invoke `ServletErrorPage`.

I'll leave this open for a few days in case I have missed something but unless we reach consensus that this test should call `ServletErrroPage` I intend to close this issue with no further action.

janbartel:
@markt-asf oops my bad. I was confused by the similarity of the error code used in the test and the error code mapped for the error page - looked like a typo. Closing.

PR #662, opened 2024-06-21
Issue #660 fix invalid error-code 

olamy:
**Issue #660 fix invalid error-code**


markt-asf:
I want to figure out why this didn't trigger a test failure when running with Tomcat.

markt-asf:
So, it looks like Tomcat's default error page happened to include the text expected by this test. Fixing that exposes a different issue in that the test's error page servlet doesn't appear to be setting one of the headers the test client is expecting.

As it looks like we are going to need a 6.1.1 TCK release, I plan to do the following:
- Update the TCK POMs in the 6.1.x branch to use the Servlet 6.1.0 API rather than 6.1.1-SNAPSHOT. In the unlikely even of a 6.1.1 API release we can update the versions accordingly at that point.
- Merge this to 6.1.x and master

markt-asf (changes requested):
I had been working on the assumption that the original bug report was correct but looking into this in more detail, I don't think it is. I am going to move the discussion back to #600 but for now I do not think this PR should be merged.

PR #670, opened 2024-07-04
Fix #615. Only API code should be in the jakarta package

### LINKED ISSUE #615: tests should not be in the jakarta package
JESP explicitly prohibits usage of `jakarta` package namespace to everything but APIs, therefore (unit) tests must be moved to a different package

markt-asf:
Move unit tests to ee.jakarta


markt-asf:
Should be uncontroversial. I'll merge tomorrow unless there are objections.

PR #680, opened 2024-07-16
Fix #525 FlushBufferTest

### LINKED ISSUE #525: [TCK Challenge] attempting to writing more than content-length bytes
**Challenged tests:**

com/sun/ts/tests/servlet/spec/httpservletresponse/URLClient.java.flushBufferTest

**TCK Version:**

servlet-tck-6.0.1

**Tested Implementation:**

Jetty

**Description:**

`Section 5.7 Closure of Response Object` states that:

```
When a response is closed, the container must immediately flush all remaining content in the response
buffer to the client. The following events indicate that the servlet has satisfied the request and that the
response object is to be closed:
• The termination of the service method of the servlet.
• The amount of content specified in the setContentLength or setContentLengthLong method of the
response has been greater than zero and has been written to the response.
• The sendError method is called.
• The sendRedirect method is called.
• The complete method on AsyncContext is called.
```

The `flushBufferTest` does:

```java
  public void flushBufferTest(HttpServletRequest request,
      HttpServletResponse response) throws ServletException, IOException {
    int size = 40;
    response.setContentLength(size);

    PrintWriter pw = response.getWriter();
    pw.println("Test PASSED");
    StringBuffer tmp = new StringBuffer(2 * size);
    int i = 0;

    while (i < 8) {
      tmp = tmp.append("111111111x");
      i = i + 1;
    }
    pw.println(tmp);
    response.addIntHeader("header1", 12345);
  }
```
The test expects a `200` response without the presence of `header1` and probably that the output has been truncated to 40 bytes.

In this case, where the servlet has attempted to write more than the `content-length` number of bytes Jetty will throw an `IOException` and return an error `500` response.

There is nothing in the wording of the spec nor javadoc that prohibits a `500` response, or that specifically requires a `200` response with excess bytes ignored. Nor would it make sense to truncate the bytes written to `content-length`, as some content encodings are mutiibyte and would result in garbage. 

If the aim of the test is to ensure that excess characters are not written, nor the `header` sent, then the test should accept an error `500` response.

 **See also:**
https://github.com/eclipse/jetty.project/issues/9651

gregw:
Replace the disputed test (that tests poorly defined behaviour of a bad servlet) with two tests that check the accepted behaviour of a well written servlet.

Once there is an accepted interpretation of the behaviour for the bad servlet, additional tests can be added to check that.

### DIFF
--- a/tck/tck-runtime/src/main/java/servlet/tck/spec/httpservletresponse/HttpTestServlet.java
+++ b/tck/tck-runtime/src/main/java/servlet/tck/spec/httpservletresponse/HttpTestServlet.java
@@ -25,7 +25,10 @@
 package servlet.tck.spec.httpservletresponse;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 
 import servlet.tck.common.servlets.HttpTCKServlet;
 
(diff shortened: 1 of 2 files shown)


olamy (approved):

tck/tck-runtime/src/main/java/servlet/tck/spec/httpservletresponse/HttpTestServlet.java
```diff
@@ -42,22 +45,44 @@ public void intHeaderTest(HttpServletRequest request,
...
+    out.write(passed);
+
+    byte[] fill = new byte[size - passed.length];
+    Arrays.fill(fill, (byte) 'x');
+    out.write(fill);
     response.addIntHeader("header1", 12345);
+
+    try {
+      out.write(fill);
+      throw new IllegalStateException("write did not fail");
```
markt-asf, line 64: Is this sufficient to trigger a test failure?

gregw: It should be as the requirement is that the proceeding write closes the output stream. Wiring to a closed output stream should be an exception regardless of buffering.

gregw: We could add an out.flush if you are really concerned, but I don't see it as necessary

markt-asf: Sorry, should have been clearer. I'm not doubting the exception will be thrown. My question is whether the client will see any effects of the exception which will in turn triggered a test failure. Or to put it another way, will the client see the completed response and think all is well regardless of whether the exception is thrown on the server.

gregw: In this test, the write a line 59 will complete the response and close the stream.  The client will see that response as a 200.     The subsequent write will throw an exception.... ah I see the problem, the client will not see the exception either way.  Let me ponder a better approach....  hard in this test environment as everything is sent via the response.

gregw: @olamy Can you test the latest version please

olamy: @gregw all good with jetty 12.1.x

gregw: @markt-asf How's this version?  It does every test request twice with any bad behaviour after response completion of the first request stored in the session to be seen by the second request.

markt-asf: Neat solution. LGTM.

gregw:
OK for me to merge this?   I've no idea what release it will end up in... but happy enough it will end up in some release some time.

olamy:
@gregw, go for it. I can then cherry-pick the `6.1.x` branch to be used for a release.

PR #756, opened 2024-11-08
Revert "Fix #525 FlushBufferTest"

### LINKED ISSUE #525: [TCK Challenge] attempting to writing more than content-length bytes

pmd1nh:
Reverts jakartaee/servlet#680

Hi,

The change in PR 680 has completely removed the old test, which has been around in many releases, and replaced with two new tests.  If the community thinks that the previous/old test is invalid, it should be excluded in the Servlet TCK 6.1.1.  Any new or replacement tests need to be added to the next servlet release.

PR 680 doesn’t follow the service release policy and needs to be reverted.

### DIFF
--- a/.github/ISSUE_TEMPLATE/tck_challenge.md
+++ b/.github/ISSUE_TEMPLATE/tck_challenge.md
@@ -1,0 +1,21 @@
+---
+name: TCK Challenge
+about: Create a TCK Challenge
+title: ''
+labels: 'challenge'
+assignees: ''
+
+---
+
+**Challenged tests**
+List the challenged tests with the fully qualified classnames and then the test methods, e.g.
+ee.jakarta.tck.faces.test.javaee7.multiFieldValidation.Spec1IT#testFailingPreconditionsNotAfterAllInputComponents
+
+**TCK Version**
+Specify the version of the TCK, e.g. Jakarta Faces 4.0.x
+
+**Description**
+A clear and concise description of why you think the tests are wrong.
+
+**Additional context**
+Add any other context about the challenge here.
--- a/.github/ISSUE_TEMPLATE/tck_challenge.md
+++ b/.github/ISSUE_TEMPLATE/tck_challenge.md
@@ -9,10 +9,10 @@ assignees: ''
 
 **Challenged tests**
 List the challenged tests with the fully qualified classnames and then the test methods, e.g.
-ee.jakarta.tck.faces.test.javaee7.multiFieldValidation.Spec1IT#testFailingPreconditionsNotAfterAllInputComponents
+servlet.tck.api.jakarta_servlet.asynccontext#dispatchZeroArgTest()
 
 **TCK Version**
-Specify the version of the TCK, e.g. Jakarta Faces 4.0.x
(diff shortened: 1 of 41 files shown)


pmd1nh:
@markt-asf @gregw @olamy

pmd1nh:
@pnicolucci

pnicolucci:
Link to the TCK process: https://jakarta.ee/committees/specification/tckprocess/

I agree with @pmd1nh  here that these changes should be reverted in the 6.1.1 Servlet TCK service release. The following challenge needs to be revisited:  https://github.com/jakartaee/servlet/issues/525.

markt-asf:
#680 is fine - it updated master. The issue is the cherry-picking of it into the 6.1.x branch in 1b90c29

markt-asf:
This PR needs to be against 6.1.x, not master.

markt-asf (changes requested):
This PR needs to be against 6.1.x rather than master.

It should also add the `@Ignore` annotation to the challenged test.

pmd1nh:
corrected branch to 6.1.X
added `@Ignore` to the `flushBufferTest()`

markt-asf (changes requested):
Note sure what you have done but there are way too many changes included.

pmd1nh:
They show up after changing to the 6.1.X.

pmd1nh:
Reference #757

PR #838, opened 2025-04-11
Fix #11 - clarify behaviour of ServletContext.getRealPath()

### LINKED ISSUE #11: Clarification required for ServletContext.getRealPath and HttpServletRequest.getPathTranslated
There appear to be different interpretations of the required behaviour of ServletContext.getRealPath(java.lang.String path) and HttpServletRequest.getPathTranslated() if the virtual path can be mapped to a physical path but the physical path does not exist.

My reading of the specification and the Javadoc is that the physical path does not have to exist, just that the mapping has to be possible. However, I can see how someone else could read the specification as requiring the physical path to exist and null to be returned if it does not.

This should be clarified for Servlet 3.1 and ideally the clarification included in any subsequent maintenance releases of the earlier specification versions.

markt-asf:
Clarification based on wording suggested by @gregw.

I'll leave this open for a week or so for comment before (assuming no objections) adding some additional TCK tests and merging.

### DIFF
--- a/api/src/main/java/jakarta/servlet/ServletContext.java
+++ b/api/src/main/java/jakarta/servlet/ServletContext.java
@@ -408,6 +408,10 @@ public interface ServletContext {
      * container is running, including the proper path separators.
      *
      * <p>
+     * The real path returned is not required to exist. Generally, that means if a non-null result is returned for the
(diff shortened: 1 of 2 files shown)


gregw (approved):

markt-asf:
Force push was to fix merge conflicts. No other changes.

markt-asf:
Just confirmed that the TCK already tests this behaviour.

PR #851, opened 2025-04-15
Fix #25 - Note risks associated with filter modifying the response

### LINKED ISSUE #25: Require that filters that modify the response body must address content length and range requests
If a filter modifies the response body it may cause problems for a servlet if the servlets set a content length or respond correctly to range requests.

The specification currently implies that wrapping the response is necessary if the response body is modified. This issue requests that filters that modify the response body are required to correctly handle (which will almost certainly mean wrapping the response) range requests and the content-length of the response.

markt-asf:
Add recommendation that such filters should minimise the risk of breakage.

I plan to leave this open for at least a week for feedback.


spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -2703,6 +2703,18 @@ resource after its call.
...
+break a subsequent filter or the servlet in the `FilterChain`. For example, if
+the servlet supports range requests or sets a content length, a filter that
+modifies the response body is likely to break the servlet. It is the
+responsibility of the filter developer to clearly describe what it does with the
+`ServletResponse` so a user of the filter is aware of what happens when they add
+the filter.
+
+Filters that modify the response body generated by a Servlet should take steps
+to ensure that they minimise the risk of breaking a subsequent filter or the
+servlet in the `FilterChain`.
```
gregw, line 2716: Do we want to be explicit and say that when modifying the body, a filter must ensure that any affected headers e.g. `Content-Length` are correctly modified as well.

gregw (approved):

spec/src/main/asciidoc/servlet-spec-body.adoc
markt-asf: That was where I started in #25 because a TCK filter was breaking Tomcat's default servlet. The argument in #25 was a filter doesn't know what a subsequent filter/servlet might do to the response.
I'm fine with being more explicit and requiring that headers are correctly modified. It will mean updating the TCK to fix the filters that don't currently do this (which in those cases I think just means clearing the content-length header). Unless there are objections, I'll add that to my TODO list.

markt-asf: No objections. Updated accordingly.

markt-asf:
I'll leave this open for another week before merging in case of any further comments.

I've tested this by reverting the Tomcat fix that addressed the original TCK issue and confirmed that the TCK fails without the TCK changes and that the TCK changes fix those failures.

PR #857, opened 2025-04-25
add TransportGuarantee.INTEGRAL

Chenjp:
add an INTEGRAL transport guarantee which is declared in jakarta-servlet-spec-6.1#13.8. Specifying Security Constraints, web-common_6_1.xsds#transport-guaranteeType


markt-asf:
This is a deliberate simplification. There are no implementations I am aware of that can provide integrity that don't also provide confidentiality (i.e. TLS) therefore the options for the annotation was kept to 2.

PR #881, opened 2025-06-12
Fix #20 - add support for <welcome-servlet>

### LINKED ISSUE #20: Clarify implicit mappings and welcome file behaviour
Consider the following:
*.do is mapped to a servlet
welcome files are index.jsp, index.do

The intention is that the index.jsp page should be used if present and index.do (which always maps to the servlet) used if it is not. However, a strict reading of the servlet spec requires that a 404 is returned if index.jsp is not present .

markt-asf:
Implements the solution discussed in #20.

I plan on leaving this open for a week or two for discussion before merging and wouldn't be surprised if changes were required.

### DIFF
--- a/spec/src/main/asciidoc/servlet-spec-body.adoc
+++ b/spec/src/main/asciidoc/servlet-spec-body.adoc
@@ -5687,23 +5687,21 @@ The web container identifies the servlet to serve the request using the process
 described in the following paragraphs. The process stops when the first match is
 found.
 
-For each welcome resource in the order it appears in the deployment descriptor:
+For each `<welcome-file>` or legacy `<welcome-file>` in the order it appears in
+the deployment descriptor:
 - append the partial URI to the request URI to create a welcome URI
-- if the welcome URI has an exact match to a servlet mapping, send the request
-  to that servlet
-- if the welcome URI has a prefix match to a servlet mapping, send the request
-  to that servlet
-- if the welcome resource is a welcome file or legacy welcome file and a static
-  file exists in the web application for the welcome URI, send the request to
-  that static resource
-- if the welcome resource is a welcome servlet and a servlet mapping (other than
-  the default servlet) matches the welcome URI, send the request to that servlet
+- if a static file exists in the web application for the welcome URI, map the
+  welcome URI using the standard servlet mapping rules and process with the
+  mapped servlet
+- if the welcome URI is an exact match to a mapping for a servlet, process with
+  that servlet
 
-If no match is found then for each legacy welcome file in the order they appear
+For each `<welcome-servlet>` or legacy `<welcome-file>` in the order it appears
 in the deployment descriptor:
 - append the partial URI to the request URI to create a welcome URI
-- if a servlet mapping (other than the default servlet) matches the welcome URI,
-  send the request to that servlet
+- map the welcome URI to a servlet using the standing mapping rules
+- if the welcome URI maps to a servlet other than the default servlet, process
+  with that servlet
 
 If a matching welcome resource is found in the manner described, the container may
 send the request to the welcome resource with a forward, a redirect, or a


markt-asf:
Re-based and made a couple clarifications to the servlet text. No functional change.

markt-asf:
Some feedback from the Tomcat community:

At the moment, a `<welcome-servlet>` is almost certain to match since it will likely be using extension mapping making any welcome resources that follow unnecessary. This might be an issue for web fragments since any welcome resources defined in a fragment will be added to the end of the existing list of welcome resources.

Is this a problem? I think probably yes, but would welcome additional views. If it is an issue, then one way to resolve it would be to change the mapping algorithm to:

For each welcome resource in the order it appears in the deployment descriptor:
- append the partial URI to the request URI to create a welcome URI
- if the welcome URI has an exact match to a servlet mapping, send the request to that servlet
- if the welcome URI has a prefix match to a servlet mapping, send the request to that servlet
- if a static file exists in the web application for the welcome URI, send the request to that static resource

If no match is found then for each legacy welcome file and `welcome-servlet` in the order they appear in the deployment descriptor:
- append the partial URI to the request URI to create a welcome URI
- if a servlet mapping (other than the default servlet) matches the welcome URI, send the request to that servlet

This does mean two passes of the welcome files are more likely but I think it is closer to the current behaviour whilst fixing the problem this change was intended to fix. I'm currently leaning towards this alternative approach.

spec/src/main/asciidoc/servlet-spec-body.adoc
```diff
@@ -5673,26 +5679,40 @@ where `directory` is an entry in the WAR that is not mapped to a
...
+The web container identifies the servlet to serve the request using the process
+described in the following paragraphs. The process stops when the first match is
+found.
+
+For each welcome resource in the order it appears in the deployment descriptor:
+- append the partial URI to the request URI to create a welcome URI
+- if the welcome URI has an exact match to a servlet mapping, send the request
+  to that servlet
+- if the welcome URI has a prefix match to a servlet mapping, send the request
+  to that servlet
```
gregw, line 5695: Why are we mapping welcome-file elements in 6.2 descriptors to servlets?   Surely these mappings should only be for `welcome-servlet` resources and perhaps legacy welcome files?

gregw: I would think that for all `welcome-file` resources, then the file must exist.  This would work for precompiled JSPs because the index.jsp file should still be there.    Then only for `welcome-servlet` resources do we look for a match regardless of the files existence?

I'm AFK until Monday, so can this wait until then for some more considered discussion?

funkman: "Long ago", it was considered a good practice after pre-compiling the jsp's to delete them to cut down on war/deployed file size. Especially helpful when you might have hundreds/thousands of jsp's. [ Contrast to today where disk is much cheaper and templating engines would probably be better then jsp's from a design choice. ] This is a long winded way of saying, enterprise installations/applications porting forward might be a common scenario. (Of deleting jsp's after pre-compile)

gregw: Sorry - I've been out. Looking now

gregw: I think I'm +0 on these changes.    I'm not exactly sure that adding <welcome-servlet> is adding much value as we are still mapping <welcome-files> to servlets without files to support backwards compatibility.

If the third bullet point is for *.jsp, then isn't that a suffix match rather than a prefix match.

To make me +1, I think  welcome-file elements should only trigger if there is a matching resource (but can be served by a servlet that also matches).  Also  I think it would be clearer to separate out the cases:

```adoc
For each welcome resource in the order it appears in the deployment descriptor:
- append the partial URI to the request URI to create a welcome URI
- if the welcome resource is a legacy welcome file:
   * if the welcome URI has an exact match to a servlet mapping, send the request  to that servlet
   * if the welcome URI has a prefix match to a servlet mapping, send the request to that servlet
   * if a static file exists in the web application for the welcome URI, send the request to the default servlet
- if the welcome resource is a welcome file:
   * if a static file exists in the web application for the welcome URI, send the request to any matching servlet including the default servlet. 
-  if the welcome resource is a welcome servlet:
   * if a servlet mapping (other than the default servlet) matches the welcome URI, send the request to that servlet

```

PR #885, opened 2025-06-27
API and specification updates to support pre-emptive authentication

### LINKED ISSUE #21: Clarify behaviour with pre-emptive authentication
The HTTP spec allows pre-emptive authentication - i.e. sending credentials before the server asks for them. It is unclear from the Servlet spec how getRemoteUser() and friends should behave in this regard when the resource being requested does not require authentication. Should the credentials be processed or ignored?

markt-asf:
Fix #21.

This is a draft PR as it only includes the API and specification changes at this point. Once consensus is reached, I'll add some TCK tests before merging.

I intend to leave this in draft for at least 2 weeks for feedback.


markt-asf:
It has been well over a month without any feedback. I'll leave it a few more days in case folks missed the original message. If I don't receive further feedback, I'll assume consensus, write some TCK tests and merge.

PR #914, opened 2025-09-08
Clarify when complete and dispatch throw an ISE

markt-asf:
This is intended only to align the Javadoc description of when an ISE is thrown with the text in the spec doc.

I'll leave this for at least a week in case others read the spec document differently.

### DIFF
--- a/api/src/main/java/jakarta/servlet/AsyncContext.java
+++ b/api/src/main/java/jakarta/servlet/AsyncContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2023 Oracle and/or its affiliates and others.
+ * Copyright (c) 2017, 2025 Oracle and/or its affiliates and others.
  * All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


markt-asf:
It has been several weeks with no feedback so I intend to merge this commit.

Before I merge I intend to make the following additional changes:
- update the specification document to add text that states `complete()` may be called no more than once per asynchronous cycle with any additional calls triggering an `IllegalStateException`
- add one or more TCK tests to check for this behaviour

I have read the specification multiple times and the text is currently ambiguous as to whether multiple calls to `complete()` per asynchronous cycle are allowed.

My reading of the specification document is that the intent was that multiple calls to complete are not allowed. This is based on the following:
- All of the start and dispatch methods have strict requirements around number and sequence of calls that are permitted. It would be consistent for `complete()` to be similarly specified. 
- The description of `complete()` includes the text "An `IllegalStateException` MUST be thrown if `startAsync` was not called.". Given that `complete()` is a method on the `AsyncContext` and the only way to obtain an `AsyncContext` is to call `startAsync()` this sentence has no purpose unless the intention was to disallow multiple calls to `complete()`.
- Request (and response) objects associated with a asynchronous cycle are only valid up to the point `complete()` is called. Asynchronous applications using multiple threads therefore need to keep track of when `complete()` is called to avoid indeterminate behaviour (see 3.14 & 5.8). "Keeping track" of when `complete()` is called is equivalent to making sure `complete()` is only called once.

PR #935, opened 2025-10-15
Update TCK to use central SNAPSHOT repo

markt-asf:


markt-asf:
> I am aware, but we don't need it here.

Ignore that. Misread your comment. I'll take a look.

PR #1026, opened 2026-03-30
do not work offline as it can prevent project downloading artifacts (#1022)

olamy:


markt-asf:
LGTM

PR #1038, opened 2026-04-30
Fix thread safety of StaticLog class

olamy:


olamy:
with this one approved https://github.com/arquillian/arquillian-core/pull/835/.
Servlet TCK run will support junit parallel execution

